PR #2, opened 2020-09-07
Support running benchmarks on a local repository

CraigFe:
Changes:

- split the code into `bin/` and `lib/` directories (to make room for making each more complicated);
- add support for passing a path to a local Git repository;
- restructure the `Cmdliner` DSL into more modular components (i.e. `Source`, `Docker_config`).
- restructure `utils.ml` slightly and add an interface

Please let me know what you think of it :slightly_smiling_face:


CraigFe:
Example local pipeline:

image

gs0510:
This is so much cleaner and modular, thank you for the refactoring! I added a commit with two very tiny corrections. :)

CraigFe:
> I added a commit with two very tiny corrections. :)

Thanks :tada:

Issue #11, opened 2020-12-14
Allow configurable Dockerfile


gs0510:
Currently, the [docker DSL](https://github.com/ocurrent/current-bench/blob/main/backend/lib/pipeline.ml#L38) specifies the steps needed to run benchmarks, but for different repositories the packages might differ, someone repositories might need pinning some packages for the benchmarks to work, hence a configurable dockerfile which can be plugged into the pipeline will be useful.

Issue #10, opened 2020-12-14
Allow configurable benchmark targets in the pipeline to allow easier enrolling.


gs0510:

rizo:
We multi-benchmark support added in #17 we can consider this done, I think. The UI work is still WIP.

PR #13, opened 2020-12-15
Allow configurable benchmark targets

rizo:
Addresses https://github.com/ocurrent/current-bench/issues/10.

Currently the assumption is made that the repository to be benchmarked has a `Makefile` with the `bench` target. It should be possible to alter this configuration for projects that follow a different existing convention.

This PR implements a simple approach to overriding the default command executed to start the benchmarks in the docker environment. This change is done via a command-line option that persists for the execution of the pipeline.

This has some limitations and presents a challenge for current-bench in deciding _when_ this should be allowed.

## Execution stages

Providing a simple user experience while working with benchmarks requires a very careful management of conventions and expectations. On the one hand our infrastructure could offer a "zero-config" service that "just works", on the other hand we could allow users to customise the different aspects of the execution of the benchmarks and the pipeline itself.

Configurability and the "zero-config" experience are not mutually exclusive. In fact, "zero-config" can be achieved with a good selection of default configuration options.

Both https://github.com/ocurrent/current-bench/issues/11 and https://github.com/ocurrent/current-bench/issues/10 add configurability to the benchmarking infrastructure. It is important to understand what this means for the benchmarking service and for the users.

From the point of view of opportunities to customise the behaviour of the benchmarks the following stages can be outlined:

1. **Pipeline initialisation time**: the main executable is started allowing to provide multiple command-line options. The provided options (normally) do not change during the execution of the pipeline. This configuration is mostly useful for pipeline instances dedicated for specific usage or for local execution.
2. **GitHub App is added to a repo**: an existing instance  of the pipeline (pre-configured as in 1) is set up for a new GitHub project to be benchmarked. When adding the app GitHub can redirect the users to a page to provide some additional configuration. On this page users could provide their preferred configuration for the app. This could include defining the benchmark target, for example.
3. **Automated execution (on commit)**: when the pipeline schedules the execution of a new benchmark, the source code of the project can contain a configuration file that can be used by the pipeline to customise the execution behaviour. This allows dynamic changes that are not possible with 1 and 2, with the commit-level granularity. For example, if users want to change the benchmarking target in a specific commit, it should be possible during this stage.
4. **Manual trigger**: works by allowing users to trigger the (re)execution of benchmarks on demand. This functionality might require some more advanced integration between the app and the repo to allow something like `/benchmark` commands in Github comments, for example. When executing the benchmark this way, users could provide additional parameters, such as the benchmarking targets. One particularly interesting use case is supporting `quick` and `slow` targets, similar to how this is implement in Alcotest.

Further analysis of requirements for simple and more complex projects is required to decide what configurations (and when!) should be possible.

## Summary

The current PR takes a conservative approach and implements the configuration to override the benchmarking command at the "Pipeline initialisation time", allowing different pipeline instances to support different targets. This might not be an ideal approach; additional usage requirements would help make a better decision.

### DIFF
--- a/backend/lib/pipeline.ml
+++ b/backend/lib/pipeline.ml
@@ -48,7 +48,7 @@ let dockerfile ~base =
   @@ run "eval $(opam env)"
 
 let default_docker_cmd =
-  "/usr/bin/setarch x86_64 --addr-no-randomize make bench"
+  [ "/usr/bin/setarch"; "x86_64"; "--addr-no-randomize"; "make"; "bench" ]
 
 let weekly = Current_cache.Schedule.v ~valid_for:(Duration.of_day 1) ()
 
@@ -85,8 +85,7 @@ let pipeline ~slack_path ~conninfo ~(info : pr_info) ~dockerfile ~cmd ~tmpfs
       @ docker_cpuset_cpus
       @ docker_cpuset_mems
     in
-    let+ output =
-      Docker.pread ~run_args image ~args:(String.split_on_char ' ' cmd)
+    let+ output = Docker.pread ~run_args image ~args:cmd
     and+ commit =
       match head with
       | `Github api_commit -> Current.return (Github.Api.Commit.hash api_commit)
--- a/backend/lib/pipeline.ml
+++ b/backend/lib/pipeline.ml
@@ -47,6 +47,9 @@ let dockerfile ~base =
   @@ add ~src:[ "--chown=opam ." ] ~dst:"." ()
   @@ run "eval $(opam env)"
 
+let default_docker_cmd =
+  "/usr/bin/setarch x86_64 --addr-no-randomize make bench"
+
 let weekly = Current_cache.Schedule.v ~valid_for:(Duration.of_day 1) ()
 
 type pr_info = [ `PR of int | `Branch of string ]
@@ -66,7 +69,7 @@ let github_status_of_state url = function
   | Error (`Active _) -> Github.Api.Status.v ~url `Pending
   | Error (`Msg m) -> Github.Api.Status.v ~url `Failure ~description:m
 
-let pipeline ~slack_path ~conninfo ~(info : pr_info) ~dockerfile ~tmpfs
+let pipeline ~slack_path ~conninfo ~(info : pr_info) ~dockerfile ~cmd ~tmpfs
     ~docker_cpuset_cpus ~docker_cpuset_mems ~head ~name ~owner =
   let src =
     match head with
@@ -83,11 +86,7 @@ let pipeline ~slack_path ~conninfo ~(info : pr_info) ~dockerfile ~tmpfs
       @ docker_cpuset_mems
     in
     let+ output =
-      Docker.pread ~run_args image
-        ~args:
-          [
-            "/usr/bin/setarch"; "x86_64"; "--addr-no-randomize"; "make"; "bench";
-          ]
+      Docker.pread ~run_args image ~args:(String.split_on_char ' ' cmd)
     and+ commit =
       match head with
       | `Github api_commit -> Current.return (Github.Api.Commit.hash api_commit)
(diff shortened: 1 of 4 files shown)


backend/lib/pipeline.ml
```diff
@@ -83,11 +86,7 @@ let pipeline ~slack_path ~conninfo ~(info : pr_info) ~dockerfile ~tmpfs
       @ docker_cpuset_mems
     in
     let+ output =
-      Docker.pread ~run_args image
-        ~args:
-          [
-            "/usr/bin/setarch"; "x86_64"; "--addr-no-randomize"; "make"; "bench";
-          ]
+      Docker.pread ~run_args image ~args:(String.split_on_char ' ' cmd)
```
MagnusS, line 89: Looks like this wouldn't support commands containing quotes. It would be useful to keep `~cmd` as `string list`, at least as input to the pipeline function. Perhaps it would be possible to also get a list from cmdliner, e.g. after a `--`?

MagnusS:
Thanks! Reading https://github.com/ocurrent/current-bench/issues/10 I'm also wondering if this solves the problem. @gs0510 could you include some more details about the use case that should be solved?

backend/lib/pipeline.ml
rizo: I agree. I also thought about this and was wondering if there's a better way to do this.

rizo: We already use a positional argument for the source (e.g. `main.exe local path_to_project`). I think it might be confusing to use `--` and positional options for the command as well.

rizo: Implemented a custom Cmdliner converter to handle parsing and printing in https://github.com/ocurrent/current-bench/pull/13/commits/27c95b6df3e912433c3005593cfc35360c847a2d. Kept the `--cmd` option.

MagnusS:
Code change LGTM - but I'm not sure about the use case so leaving open for @gs0510's review. With a command line option it's still static and the same for all repos monitored by the pipeline.

gs0510:
@MagnusS  The use case I was thinking of was like that of `irmin`'s at the moment. Only `bench layers` can be run at the moment since that's what is in a parse-able format for the pipeline. I know we want zero configuration but it seems that it'll be easier to get repositories in the pipeline if this  target can be configurable.

rizo:
@gs0510 Thanks for clarifying this! I think with configurable benchmark targets we would still only support one target at a time. The users would have to change the configuration if they wanted to run a different target. I think this means we would need a per-commit configuration (point 3 in my first post). But even in that case, how is it different to changing what `make bench` runs in a given commit?

There's another way... If the problem is to support multiple targets for the same project, we could implement the ability to combine results from multiple benchmark runs. AFAIU, currently the output of the `docker run ... make bench` is captured and directly processed. This creates a 1-1 relation between the benchmark target and the pipeline. I think an alternative could be to have `make bench` write one or more JSON benchmark files to a specified directory. With this approach the pipeline would read all of the produced files (for different benchmark runs all triggered by a single `make bench`) and combine them into a meaningful internal representation.

It is my understanding that running benchmarks for the `layers` target reliably is higher priority than supporting multiple targets. I don't think the proposed implementation is sufficient to fully support multiple targets, but we can keep this PR open to discuss the design requirements.

rizo:
One elaboration to my suggestion that does not require almost any changes to the output is supporting parsing of multiple JSON objects in the output of `make bench`.

What I'm imagining is a `make bench` target like:

```make
bench:
    ./benchmark1
    ./benchmark2
```

Both of the benchmarks sequentially print the currently defined JSON output and the pipeline processes and combines them.

Let me know if this is a bad idea :)

gs0510:
Adding the ability to process multiple jsons is a better idea than having configurable benchmarks! I realized I am wrong in thinking that having configurable targets (the way I was envisioning it was having a target for `irmin`, having another for `index` etc. etc)  would make this problem go away since we will also need new docker run for every target.

MagnusS:
I think supporting multiple groups of json output would be useful.

rizo:
Closing in favour of #17

PR #17, opened 2020-12-21
Read multiple JSON objects from benchmark output.

rizo:
Implementation of https://github.com/ocurrent/current-bench/pull/13#issuecomment-746279444.

The changes to support multiple JSON objects in the benchmark output is pretty simple. The problematic part is knowing how to merge the multiple results. The current SQL schema disallows multiple entries per commit. I think there are two solutions to this:

1. Add an additional column to the SQL schema for benchmarks table. This would represent the unique name of the benchmark for a given repo and commit. The JSON format should be updated to have a top-level `name` key in this case.
2. Merge the individual entries under `results` in the JSON output and only insert one results object into the database per commit. This does not require changes to the JSON output format, but requires that the result "names" are unique across different benchmarks.

I'm leaning towards supporting multiple benchmarks per repo/commit, i.e., 1.

### DIFF
--- a/backend/postgres/init.sql
+++ b/backend/postgres/init.sql
@@ -2,7 +2,7 @@ GRANT ALL PRIVILEGES ON DATABASE docker TO docker;
 
 CREATE TABLE benchmarks(
 	repositories varchar(256),
-	commits varchar(50) NOT NULL UNIQUE,
+	commits varchar(50) NOT NULL,
 	json_data jsonb,
 	timestamp float8,
 	branch varchar(256)
--- a/README.md
+++ b/README.md
@@ -69,13 +69,13 @@ You can find more options for different configurations, posting message to slack
 If you want to enroll your repository or setup this benchmark repository for your repository,
 we make the following assumptions.
 
-1. There is a dune bench target which can run the benchmarks.
+1. There is a `make bench` target which can run the benchmarks.
 2. The benchmarks results are json with of the following format:
 ```
 {
   "results" : [
     {
-      "name": <name-of-the-benchmarks>,
+      "name": <name-of-the-test>,
       "metrics": {
         "<metric-1>": "",
         "<metric2>": "",
@@ -86,6 +86,9 @@ we make the following assumptions.
   ]
 }
 ```
+
+> Note: multiple concatenated JSON objects can be produced and will be interpreted as different benchmarks. The names of the benchmark test must be unique across the benchmark results.
+
 [Here's](https://gist.github.com/gs0510/9ef5d47582b7fbf8dda6df0af08537e4) an example from [index](https://github.com/mirage/index) with regards to what the format looks like.
 
 The metadata about `repo`, `branch` and `commit` is added by the pipeline. Your repo only needs to output the results as a json.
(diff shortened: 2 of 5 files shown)


rizo:
This is now ready for review.

The implemented approach is a mixture of 1 and 2 described above. After removing the `UNIQUE` constraint on `commits` we can have multiple benchmark entries without conflicts. An important consideration is that the under `results` must be unique across different benchmarks.

rizo:
As a practical example I tested this multi-object output:

```json
{
  "results": [
    {
      "name": "bench_1_test_1",
      "metrics": {
        "time": 0.02,
        "ops_per_sec": 100.0,
        "mbs_per_sec": 5.0
      }
    },
    {
      "name": "bench_1_test_2",
      "metrics": {
        "time": 0.04,
        "ops_per_sec": 455.0,
        "mbs_per_sec": 7.0
      }
    }
  ]
}
{
  "results": [
    {
      "name": "bench_2_test_1",
      "metrics": {
        "time": 0.01,
        "ops_per_sec": 50.0,
        "mbs_per_sec": 3.0
      }
    },
    {
      "name": "bench_2_test_2",
      "metrics": {
        "time": 0.07,
        "ops_per_sec": 877.0,
        "mbs_per_sec": 23.0
      }
    }
  ]
}
```

This results in the following entries in the database:

```
select * from benchmarksrun;
                 commits                  |      name      | time | ops_per_sec | mbs_per_sec | timestamp  |      branch      
------------------------------------------+----------------+------+-------------+-------------+------------+------------------
 f633b36cb34216d6eab2fb9aceaa1e810c24afd5 | bench_1_test_1 | 0.02 |         100 |           5 | 1608645637 | local/local/HEAD
 f633b36cb34216d6eab2fb9aceaa1e810c24afd5 | bench_1_test_2 | 0.04 |         455 |           7 | 1608645637 | local/local/HEAD
 f633b36cb34216d6eab2fb9aceaa1e810c24afd5 | bench_2_test_1 | 0.01 |          50 |           3 | 1608645637 | local/local/HEAD
 f633b36cb34216d6eab2fb9aceaa1e810c24afd5 | bench_2_test_2 | 0.07 |         877 |          23 | 1608645637 | local/local/HEAD
(4 rows)
```

With a graph for each entry in this table. For example:

Screenshot 2020-12-22 at 14 04 49

backend/postgres/init.sql
```diff
@@ -2,7 +2,7 @@ GRANT ALL PRIVILEGES ON DATABASE docker TO docker;
 
 CREATE TABLE benchmarks(
 	repositories varchar(256),
-	commits varchar(50) NOT NULL UNIQUE,
+	commits varchar(50) NOT NULL,
```
MagnusS, line 5: If repo+commit is required to be unique, it may also be useful to have a UNIQUE(repositories, commits). Should branch also be included in this?

rizo: My suggestion is that we don't actually want to have a unique constraint on repo+commit columns.

It is IMO useful to be able to re-run the benchmarks for the same commit multiple times to increase confidence in the results. The repeated runs could be used to calculate more advanced stats. For now just showing the repeated runs on the graph with the same commit label isn't very problematic.

I plan on improving the database schema (to remove the duplication between `benchmarks` and `benchmarksrun`) in future PRs where I can look into this more carefully.

MagnusS: Yes, that makes sense - thanks

gs0510:
LGTM! I think we can have multiple benchmarks for now, and think about how to capture confidence values later.

Issue #16, opened 2020-12-21
Support graph descriptions


MagnusS:
Since all the graphs are displayed in the same way right now it could be useful with support for a description of each graph that is displayed below the title in the UI. It could then be possible to include additional information, such as a short explanation of what the graph shows, higher/lower is better etc.

Issue #14, opened 2020-12-18
Show benchmarks logs


samoht:
It's useful to see the benchmarks logs when the benchmarks are running or failed (or just if someone wants to reproduce the results and see what have been built). Currently these logs are not accessible.

gs0510:
Fixed by #20 You can now see the logs on the frontend.

PR #20, opened 2020-12-30
Add new frontend

rizo:
This PR implements a new ReScript/BuckleScript frontend for benchmark visualisation. Apologies in advance for a massive diff.

The reason I decided to implement a new frontend are:
- I am more familiar with the BuckleScript stack which contributes to my productivity.
- Having types on the front-end is very nice and leads to a better development experience. This should help us as the project grows.
- I'm hoping to reuse the visual building blocks defined for benchmarks for other projects such as the Irmin daemon.

### Changes

1. Added a new `frontend-new` project alongside the existing `frontend`. I suggest to keep both while the new frontend is being tested and improved.
2. Implemented GraphQL queries for the new `benchmarks` database table. This resolves an issue that exists in the current frontend implementation where the metrics are hardcoded.
3. Added two UI helper modules:
    - `Sx` (Styling Extensions) - for utility-based CSS definitions. This is inspired by [TailwindCSS](https://tailwindcss.com).
    - `Components` - a small component library built on top of `Sx`.
    - My goal is to extract these into a separate library in the future.
4. The new frontend makes several changes to the way the benchmark graphs are laid out:
    - Each benchmark test gets its own section and not just a single graph.
    - Multiple benchmark metrics are rendered as separate graphs. This is mostly motivated by readability and the fact that the time series have different units. It should be possible to overlay multiple graphs if there're requirements for that.
    - The x-axis presents commits over time. Each commit label is a link that goes to the respective commit.
    - The graphs limit the rendered data to 20 commits, this is easy to change but I'll add support for a date-picker in the future.

Currently the frontend is configured to make requests directly to `http://autumn.ocamllabs.io:8080/v1/graphql`. To start the new frontend locally install the dependencies with `yarn` and then `yarn start`. Note that the GraphQL secret has to be set in `index.res` first.

Here's how the new frontend looks:

Screenshot 2020-12-30 at 11 46 20


samoht:
That looks really great! Just a question regarding the "20 commits". I think it would be useful to show the associated tags too (e.g. `master` or `pr/234`) and/or to add the ability to compare with a specific tag or a specific commit. Not totally sure how to display this, but the use-case is: I want to know how much the current PR improve things vs. the current master, and using only commits there is often not very practical.

gs0510:
This looks great to me too! Will we have different page views for different PRs (as we have in the old front-end)?

rizo:
> Just a question regarding the "20 commits". I think it would be useful to show the associated tags too (e.g. master or pr/234) and/or to add the ability to compare with a specific tag or a specific commit. Not totally sure how to display this, but the use-case is: I want to know how much the current PR improve things vs. the current master (or vs. an other PR), and using only commits there is often not very practical.

I definitely agree with this. The current view is centred around individual commits, but those should be grouped as PRs and be compared to master by default, as you explained.

I want to make these changes incrementally, starting with feature parity with the existing UI. I'll likely make some quick mock ups of the UI before implementing them to validate the use cases.

rizo:
> Will we have different page views for different PRs (as we have in the old front-end)?

Yes! Do you think that each PR view should only render commits for that particular PR? It seems like the current version renders all parent commits in addition to that?

gs0510:
Rendering the parent commits makes it's easier to compare the PR results to the master so I think that's a good idea!

gs0510:
Merging this since it looks good! We can keep working incrementally on the changes :)

rizo:
Thanks @gs0510!

I did some UI/UX exploration to better understand what exactly we want to present and how. I think this is important to satisfy the different requirements we have for the project. A significant part of the project is user-facing and requires interaction with graphs for basic analysis and a little bit more complex comparisons (i.e., compare a PR to to master or another commit, as suggested by @samoht).

Here are some of the early drafts I created. Things are a bit rough and incomplete, but I left annotations to explain what it all means. There are three main views: **Index View** (shows the master metrics), **PR View** and **Commit View**.

## Master View

OCaml Benchmarks - Master View

#### Annotations

1. **Benchmarks menu** - this is only rendered if the project at some point in time produced more than one benchmark results output (as implemented in https://github.com/ocurrent/current-bench/pull/17). For example, we have multiple benchmarks for `mirage/irmin`, namely `mem`, `layers` and `pack` – these would be presented as separate items in the menu.
2. **Pull requests menu** - all of the tracked PRs are listed here. This is identical to what we already have in http://autumn.ocamllabs.io:3030/prs. Links in this menu takes us to the "PR View".
3. **Latest benchmark configuration** - the benchmark output contains a `config` keys which is currently being ignored. This table simply renders the configuration values for the last commit on master.
4. **Metric graphs** - all of the metric graphs for each benchmark test are rendered similar separately with a some extra bits of info: test name, value for the last commit, unit (this is currently not included in the JSON output, but should be easy to add), the percentage delta compared to previous commit. Because this view is showing metrics just for `master`, I wasn't sure if there's anything more useful. The PR view improves this. Also see 6.
5. **Date picker component** - this allows to select a specific time interval to be rendered. Currently we are fetching everything, which is already becoming quite slow. By default the last two weeks (with data) should be fetched.
6. Settings menu - this contains some basic visualisation options: the currently implemented "graph synchronisation" (see https://dygraphs.com/gallery/#g/synchronize for a demonstration); and an option to control the graph values, which I'm considering to implement:
    - Absolute values - shows the raw values as we currently do;
    - Relative to previous - for each datapoint this is a percentage compared to the previous commit.
    - Relative to latest master - for each datapoint this is a percentage compared to the last commit in master. I think this one will be particularly useful for PR views.
7. **Test info tooltip** - a description of the test and the metrics. Related to https://github.com/ocurrent/current-bench/issues/16.


## PR View

OCaml Benchmarks - PR View

#### Annotations

1. **"Back to Pull Requests" link** - takes us back to "Master View".
2. **Commits menu** - the list of all tracked commits for the selected pull request. Links in this menu take us to the "Commit View".
3. **Pull request title** - shows the current title of the PR and has a link to the associated GitHub PR.
4. **Parent commits area** - the parent commits are greyed out to let the user focus on the commits for the selected PR. This is related to what you [said](https://github.com/ocurrent/current-bench/pull/20#issuecomment-754041491) @gs0510.


## Commit View

OCaml Benchmarks - Commit View

#### Annotations

1. **"Back to Pull Request" link** - takes us back to the parent pull request where the commit was selected.
2. **Commit menu** - contains other useful things related to the commit, such as logs (as requested in https://github.com/ocurrent/current-bench/issues/14) and the dockerfile used to run the benchmarks (this might be specially useful when/if we implement https://github.com/ocurrent/current-bench/issues/11).
3. **Commit benchmark configuration** - similarly to 3 in "Master View", this shows the `config` for a given commit.
4. **Metric value cards** - since we are looking at a single commit, there isn't a time-series to render as a graph. A simple value is presented with a comparison indicator relative to master (see 5).
5. **Comparison branch selector** - the presented commit can be compared against master or any other tag. This is rendered as a percentage in 4. Showing the absolute value for the secondary commit could also be useful.


That's it. Hope this makes sense, let me know if something isn't clear. I think having mocks that encompass the features that we want to see implemented will help us a lot with data modelling and development going forward.

Did I forget something? Did I get something wrong?

gs0510:
This is brilliant @rizo!! I think the config values shouldn't change ideally so maybe we can display it in a sidebar? (i don't have very strong feelings either way).

PR #22, opened 2021-01-19
New db schema and migrations

rizo:
In parallel to the UI changes I started updating the database model to ensure we have the data that we need to render the UI.

This PR proposes the following schema for benchmarks:

```sql
CREATE TABLE benchmarks(
	run_at timestamp without time zone,
	duration interval NOT NULL,
	repo_id varchar(256) NOT NULL,
	commit varchar(50) NOT NULL,
	branch varchar(256),
	pull_number integer,
	name varchar(256),
	data jsonb
);
```

In addition to this I added `omigrate` as a development dependency to the project to help with future schema changes.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,0 +1,18 @@
+DB_MIGRATIONS_DIR="./db/migrations"
+DB_URI="postgresql://docker:docker@localhost:5432/docker"
+
+.PHONY: db-migrate-setup
+db-migrate-setup:
+	omigrate setup --verbose --source=${DB_MIGRATIONS_DIR} --database=${DB_URI}
+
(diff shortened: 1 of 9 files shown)


gs0510 (approved):
LGTM @rizo! Are you adding a duration to check how stable the becnhmarks are? Will it be displayed in the graphs?

rizo:
I think knowing the duration is generally useful. Specially because we want to add quick/slow annotations. This would allow us to better quantify quick and slow.

And yes, I think having a graph for this would be good too. This would be an extra metric provided by OCaml Benchmarks. I discussed with Ewan adding some other metrics like overall memory/cpu usage during the benchmark.

PR #24, opened 2021-01-26
Minor deployment fixes

rizo:
Preparatory fixes for deployment.

My plan is to allow starting the entire current-bench infrastructure with a single `systemctl start current-bench` on our deployment machine.

This PR:

- Adds the systemd init file;
- Fixes a volume mount to use the docker volume instead of a local folder;
- Adds the nginx config file from the old frontend;
- Changes the port number for the UI from 82 to 80 inside the container and from 3000 to 80 externally.

I'm planning on creating a `current-bench` user on `autumn` to have a dedicated environment for deployments. Any objections to that?

### DIFF
--- a/backend/current-bench.service
+++ b/backend/current-bench.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=OCaml Benchmarks Service
+Requires=docker.service
+After=docker.service
+
+[Service]
+User=current-bench
+Type=oneshot
+RemainAfterExit=true
+WorkingDirectory=/home/current-bench/
+ExecStart=/usr/bin/docker-compose up -d --remove-orphans
+ExecStop=/usr/bin/docker-compose down
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


rizo:
@gs0510 the docker volume fix requires changes to the currently deployed services.

Here's the full explanation for clarity:

- Using `./db_data` in `volumes` instead of just `db_data` will mount the `./db_data` folder and not the `db_data` volume (created at the end of our `docker-compose` file).
- I switched this to use the actual `db_data` volume.
- Now the problem is that the production deployment is storing data under `/data/gargi/index-benchmarks/db_data`.
- We also had an empty docker volume (try `docker volume list`).
- I manually copied all files from `/data/gargi/index-benchmarks/db_data` to the the `db_data` volume.

There's one thing missing now to make sure we safely transition to the volume: we need to restart the compose services.

gs0510:
> I'm planning on creating a current-bench user on autumn to have a dedicated environment for deployments. Any objections to that?
Should be okay! 

> There's one thing missing now to make sure we safely transition to the volume: we need to restart the compose services.
Can restart without issues since it'll pick the data up from `db-data`.

PR #30, opened 2021-02-01
adapt backend with recent changes

tatchi:
I encountered some issues due to the recent db changes while trying to run the `pipeline` on a local repository.  With the changes from this PR, the `pipeline` is running without errors and database is populated with the benchmark data.

### DIFF
--- a/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
+++ b/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE benchmarks ADD duration interval;
--- a/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
+++ b/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
@@ -1,1 +1,0 @@
-ALTER TABLE benchmarks ADD duration interval NOT NULL;
--- a/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
+++ b/backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE benchmarks ADD duration interval NOT NULL;
--- a/backend/lib/models.ml
+++ b/backend/lib/models.ml
@@ -3,7 +3,6 @@ open Utils
 module Benchmark = struct
   type t = {
     run_at : Ptime.t;
-    duration : Ptime.span;
     repo_id : string * string;
     commit : string;
     branch : string option;
@@ -13,12 +12,11 @@ module Benchmark = struct
     metrics : Yojson.Safe.t;
   }
 
-  let make ~run_at ~duration ~repo_id ~commit ?branch ?pull_number data =
+  let make ~run_at ~repo_id ~commit ?branch ?pull_number data =
     let test_name = Yojson.Safe.Util.(member "name" data |> to_string) in
     let metrics = Yojson.Safe.Util.(member "metrics" data) in
     {
       run_at;
-      duration;
       repo_id;
       commit;
       branch;
@@ -29,8 +27,6 @@ module Benchmark = struct
     }
 
   let run_at self = self.run_at
-
-  let duration self = self.duration
 
   let repo_id self = self.repo_id
 
@@ -51,8 +47,7 @@ module Benchmark = struct
     record
       [
         field "run_at" run_at Ptime.pp;
-        field "duration" duration Ptime.Span.pp;
-        field "repo_id" duration Ptime.Span.pp;
+        field "repo_id" repo_id Fmt.(pair ~sep:(Fmt.any "/") string string);
         field "commit" commit Fmt.string;
         field "branch" branch Fmt.(option string);
         field "pull_number" pull_number Fmt.(option int);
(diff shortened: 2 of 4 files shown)


backend/lib/models.ml
```diff
@@ -8,13 +8,25 @@ module Benchmark = struct
...
+    let test_name = Yojson.Safe.Util.(member "name" data |> to_string) in
+    let metrics = Yojson.Safe.Util.(member "metrics" data) in
+    {
+      run_at;
+      duration;
+      repo_id;
+      commit;
+      branch;
+      pull_number;
+      benchmark_name = None;
```
tatchi, line 26: I don't know where we should grab this data from? I didn't see any `benchmark_name` field in the benchmark results..

backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE benchmarks ADD duration interval NOT NULL;
```
tatchi, line 1: This column was removed in #25.

backend/lib/models.ml
rizo: This can be obtained from the optional top-level `name` field. Tried to clarify this in the docs: https://github.com/ocurrent/current-bench/commit/805c8bc0bad18ea704d4bd14a567bfa080644639.

backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
rizo: I'd prefer if we dropped `duration`. I thought it would be useful but later realised that mirage/index benchmarks already compute the duration (as the "time" value under "metrics).

backend/lib/pipeline.ml
```diff
@@ -95,7 +94,9 @@ let pipeline ~slack_path ~conninfo ?branch ?pull_number ~dockerfile ~tmpfs
     let () =
       let db = new Postgresql.connection ~conninfo () in
       output
-      |> Json_utils.parse_many
+      |> Yojson.Safe.from_string
```
rizo, line 97: The `result` member parsing here looks good, but the ability to support multiple objects is removed.

Essentially what we want is to:

1. Parse "many" JSON objects.
2. Extract "name" from the root of the object as ~benchmark_name.
3. Iterate on "results" from each object and make benchmark values.

rizo (changes requested):
Thanks for submitting this. I've been using historical data and forgot this was broken. See my comment on the `duration` column and the multi-object parsing.

backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
rizo: I'd prefer if we dropped `duration`. I thought it would be useful but later realised benchmarks like mirage/index compute their own duration (included in the "metrics" as "time").

gs0510: The duration might still be useful to figure out how stable the benchmarks are on autumn, we might not necessarily display it, but it can be used to figure out if the machine has to be optimised further.

backend/lib/models.ml
```diff
@@ -28,43 +41,58 @@ module Benchmark = struct
...
+  let metrics self = self.metrics
 
   let pp =
     let open Fmt.Dump in
     record
       [
         field "run_at" run_at Ptime.pp;
         field "duration" duration Ptime.Span.pp;
-        field "repo_id" duration Ptime.Span.pp;
+        field "repo_id" repo_id Fmt.(pair ~sep:(Fmt.any "/") string string);
```
tatchi, line 56: There was a slight mistake actually here

backend/lib/pipeline.ml
tatchi: Thanks, I adapted the parsing code :)

backend/db/migrations/20210101173805_add_duration_column_to_benchmarks.up.sql
MagnusS: Usually the benchmarks will have a loop that they measure the duration of - the duration of the full benchmark (ie including init and shutdown) may not be designed to be stable, even if the main benchmark result is. So I’m not sure this value will be that useful to track if the benchmark itself doesn’t report it

rizo: @tatchi There's no need to delete this column because it doesn't exist in production. The CI to apply migration's hasn't been set up yet, so I've been making the changes manually.

At this stage, given our priorities, I think the benchmark authors should measure the execution duration of the full benchmark themselves if they find it useful.

We can always add this later, I'm just not happy with having the `duration` column now because:

- It's not immediately useful.
- We might want to add other generic metrics such as container CPU and mem usage. This should be probably stored inside the `metrics` column with a special prefix such as `ocaml-benchmarks.cpu`, or something along these lines.

backend/lib/models.ml
rizo: Nice catch! I haven't used the printing functions yet so didn't notice the problem.

rizo (approved):
Thanks for the changes. Looks good now.

I still think we don't need `duration` but don't feel very strongly about it. If others think we should have it, let's keep it.

PR #36, opened 2021-02-10
change default frontend port

tatchi:
To avoid conflicting with the backend (hasura) which is also `8080`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "reason-react": "^0.9.1",
-    "vite": "^2.0.3"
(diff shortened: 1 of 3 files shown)


gs0510:
On autumn, port 8081 is used by the pipeline, so maybe we can use another port :grimacing: The docker-compose file here specifies the port as 80 for the frontend: https://github.com/ocurrent/current-bench/blob/main/docker-compose.yaml#L28

Maybe we can use another default here for local testing?

rizo:
How about the following mapping?

| Service | Production port | Development port |
|---|---|---|
| Pipeline | 8081 | 8081 |
| Hasura | 8080 | 8080 |
| Frontend | 80 | 8082 |

@tatchi, If this works, could you update the PR?

PR #43, opened 2021-02-23
Better UI errors

rizo:
Adds a welcome screen and a better UI for errors:

Screenshot 2021-02-23 at 12 37 58
Screenshot 2021-02-23 at 12 38 39

### DIFF
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -144,6 +144,42 @@ let getDefaultDateRange = {
   }
 }
 
+module Welcome = {
+  @react.component
+  let make = () => {
+    <Column alignX=#center sx=[Sx.mt.xl]>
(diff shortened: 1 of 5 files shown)


gs0510 (approved):
Looks good to me! Maybe the dropdown selector can also have search in case there are a bunch of repositories in the future. 

The other question I have is when will the users see the error page(other than entering the URL wrong), maybe we should redirect to the error page from the backend also.

rizo:
Yup, once we get many repos, we'll need to change the dropdown to make it more manageable.

The error page is indeed (currently) for cases when an incorrect URL is used. But it also checks if the repo exists. For example http://autumn.ocamllabs.io/rizo/fake-repo producer an error `No such repository: rizo/fake-repo
`.

Issue #39, opened 2021-02-22
Investigate why there are repeated entries for the same commit


rizo:
And decide what to do with the repeated entries.

gs0510:
The repeated entry exists because currently the docker pull is scheduled to be daily: https://github.com/ocurrent/current-bench/blob/main/backend/lib/pipeline.ml#L53 (so that we get the metric for the master daily) but since there isn't a new commit everyday on the master we have multiple entries.  We could change it to weekly, but the duplicates might still exist and we can `update` instead of `add` to the table.

PR #44, opened 2021-02-23
Fix master ordering error

rizo:
Fixes a with master data for comparison. Instead of requesting the last N commits we were requesting the first N commits. I couldn't find a clean solution for this with Hasura. The current approach reverses the obtained commits before using them.

In addition to that, I increased the limit from 10 to 50. This is because 10 is _not_ the number of commits, but the number of datapoints across different test_cases. I'm not sure what's an ideal number for this, but we can always change it later.

### DIFF
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -27,7 +27,7 @@ query ($repoId: String!, $pullNumber: Int, $isMaster: Boolean!, $startDate: time
     ...BenchmarkMetrics
   }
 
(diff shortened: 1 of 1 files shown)


rizo:
@gs0510 I agree we should decide what to do with those. I'm not certain that's the best approach here.

We have a dedicated issue for this: https://github.com/ocurrent/current-bench/issues/39.

Any objections to the current PR apart from this issue?

gs0510:
No looks good!

PR #45, opened 2021-02-23
Fix incorrect annotation position.

rizo:
The annotation is currently pointing at the last master/main commit instead of the first pull request commit. This PR fixes that.


gs0510:
LGTM! Thanks @rizo!

PR #46, opened 2021-02-23
avoid array mutation

tatchi:
I'm experiencing some strange behavior when I switch from one PR to another one (graphs change multiple times). The array mutation introduced [here](https://github.com/ocurrent/current-bench/pull/44/files#diff-8f4030cb251d01d332725cad16d13873cbd7fd43197683c0c0a52fac75fe1281R116) seems to be the cause of the issue.


rizo (approved):
Thanks for the fix.

I guess this is happening because `data.comparisonBenchmarks` is reused between draws. Which means there's an opportunity caching the reversal operation in the future if we want.

PR #12, opened 2020-12-15
Allow configurable dockerfile

rizo:
Addresses https://github.com/ocurrent/current-bench/issues/11.

### Changes
1. The an optional `dockerfile: Fpath.t` argument is added to `Pipeline.v`.
    - If provided, the pipeline will attempt to build a custom docker image from the referenced file.
    - If not provided, the existing functionality is used, ,i.e., use the default dockerfile for opam projects.
2. The main binary is updated to expose the `--dockerfile` option.
3. The README is updated to describe the new functionality.

PR #52, opened 2021-03-04
Add git-lfs for irmin tree benchmarks

gs0510:


rizo:
Happy to merge this as a workaround. But we should not add all project dependencies to the base dockerfile. I'm currently working on a version of https://github.com/ocurrent/current-bench/pull/12 that allows project to have a `bench.docker` file which will be used to build and run the project.

Issue #15, opened 2020-12-18
Ensure the opam repository is up to date before building the project


rizo:
The docker image produced by current-bench to run the benchmarks might not be using an up to date opam repository.

This can be causing errors during package installation and terminating the benchmark execution with "Docker build exited with status 20".

@samoht
> I think the base ocaml-ci images are tracking opam-repository and trigger a rebuild whenever a new version of a dependency is merged in opam-repository. Would be nice if ocaml-bench could do something similar?

avsm:
You can actually manually override that 'schedule:weekly' by clicking on the `pull` box from the UI and clicking rebuild. It's not ideal, but a useful workaround for when there are decoupled components. (e.g. we do this in opam-repo-ci when we push new trunk images of the compiler and want something tested)

rizo:
Thanks for the tip, @avsm! Unfortunately it seems like the base image is out of date in this case, so `pull` didn't help.

rizo:
It seems that we were using an incorrect base image (`ocurrent/opam` instead of `ocaml/opam`). But even in that case, the images are only rebuilt weekly as suggested by https://github.com/ocurrent/docker-base-images/blob/25f48d7ca661bc43ef53e8dc2b396fcc30928855/src/pipeline.ml#L8.

gs0510:
This is fixed by #55

PR #55, opened 2021-03-09
Docker environments and other improvements

rizo:
Apologies in advance for a massive diff. This change required a pretty massive refactoring of the project so I took this opportunity to clean up some things.

Below is the overview of the changes.

### Dockerize everything

The pipeline can now be built and run in a docker container. This is achieved by sharing `/var/docker.sock` into the container to allow redirecting any docker CLI commands to the host.

I added the pipeline to the docker-compose too and as a result the entire project can be now started (locally) with a single `make start-development`.

### Introduce development and production environments

The `./environments` folder contains docker-compose files for the development and production deployments. They can have different configurations and different services.

The development environment runs locally and automatically applies database migrations. It works with a local git repository that should allow for very easy testing.

The production environment requires migrations to be run by manually. This is intentional to avoid any risky automation. And of course, the pipeline runs in the `github_app` mode.

There are `.env` files for both the development and production environments. This should make it easy for us to set secrets and custom execution options.

### Improve logging

When running the pipeline with `--verbosity=debug` the whole universe prints debug messages about the everything that's happening in the app (it is very very noisy). Our logging module now filters out third-party sources and applies the warning level to them even if your app is in the debug verbosity.

### Add references to build and run job ids

This can be later queried by the frontend to generate links to the logs page. This change is not related to docker but I wanted to test the migrations in the docker-compose context so adding this made things easier.

### Change the base image from `ocurrent/opam` to `ocaml/opam`

Related to https://github.com/ocurrent/current-bench/issues/15. I thought we had made this change already, but maybe we forgot.

### Add `HACKING.md`

With instructions on how to configure and start the development environment. Any feedback on this is very appreciated.

### Add a repo for local testing

This change ensures that we can have a reproducible testing repo within `current-bench` itself (more details in `HACKING.md`).

### Remove unused code in backend

Removes: the old Makefile, the old frontend, systemd init files, the old init sql files and unused utils.

### Add wait-for script

It is a common workaround to use "wait-for" scripts that await for a docker compose service to start before starting another service. More info: https://docs.docker.com/compose/startup-order/

I added a wait-for script to be used with the migrations service. We want to make sure that the pg database is ready before running the migrations.


environments/development.env.template
```diff
@@ -0,0 +1,4 @@
+OCAML_BENCH_DOCKER_CPU=1
+OCAML_BENCH_DB_PASSWORD=docker
+OCAML_BENCH_GRAPHQL_KEY=secret
+OCAML_BENCH_TARGET_ARCH=arm64
```
tatchi, line 4:
```suggestion
OCAML_BENCH_TARGET_ARCH=amd64
```

rizo: Good catch! I'm actually on arm, forgot to switch this back.

environments/development.docker-compose.yaml
```diff
@@ -0,0 +1,66 @@
...
+    restart: always
+    environment:
+      HASURA_GRAPHQL_DATABASE_URL: postgres://docker:docker@db:5432/docker
+      HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
+      HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
+      HASURA_GRAPHQL_ADMIN_SECRET: ${OCAML_BENCH_GRAPHQL_KEY?required}
+  frontend:
+    build: ../frontend
+    ports:
+    - "80:8082"
```
tatchi, line 26:
```suggestion
    - "8082:80"
```

rizo: Yes! `HOST:CONTAINER`.

I had some issues compiling the backend on arm64 so couldn't test it. It seems to work fine on the production machine though.

I also had issues with hasura constantly running out of mem and being OOM killed by docker.

I suspected that these are arm64 specific.

If you run into any issues while starting compose, please report them.

tatchi:
I didn't read the code but followed the `HACKING.md` instructions and everything went fine. That's very nice to have all this stuff dockerized, amazing work! 

I only have 2 suggestions:

* Might be worth adding `*.env` in `.gitignore` so that when people create their env file from the template, it doesn't get mistakenly committed.
* When I started the `frontend`, I was a bit surprised to actually see data from `http://autumn.ocamllabs.io`. That would be nice to grab the backend URL from env (it's currently hardcoded in the code). Vite [allows reading `.env` files](https://vitejs.dev/guide/env-and-mode.html#env-files) but from what I understand, they should be located at the root of the project (within `frontend/` in our case). Maybe we could copy (and rename) the env file e.g `development.env` within the `frontend` folder during one of the docker steps?  Not sure though if that's a good idea 😅

README.md
```diff
@@ -1,85 +1,24 @@
...
-### Benchmarks format
-
-If you want to enroll your repository or setup this benchmark repository for your repository,
-we make the following assumptions.
+If you want to enroll your repository or setup this benchmark repository for your repository, we make the following assumptions.
 
 1. There is a `make bench` target which can run the benchmarks.
-2. The benchmarks results are json with of the following format:
-```
+2. The benchmarks result is a JSON object with of the following format:
```
tatchi, line 10:
```suggestion
2. The benchmarks result is a JSON object of the following format:
```

environments/production.docker-compose.yaml
```diff
@@ -0,0 +1,57 @@
+version: '3.6'
```
tatchi, line 1: I don't know if it matters but `development.docker-compose.yaml` uses version `3.8`

README.md
```diff
@@ -88,38 +27,24 @@ we make the following assumptions.
...
-Follow instructions from instructions here: https://hasura.io/docs/1.0/graphql/manual/deployment/docker/index.html
+### Multiple benchmarks per project
 
-### Troubleshooting
+Multiple concatenated JSON objects can be produced and will be interpreted as different benchmarks. The name of the benchmark is optional when there is only one output, but must be present if multiple result objects are produced.
 
-You can `export FRONTEND_DIRECTORY=<location of frontend directory>` as a env variable if it is ignored by `docker-compose build`.
 
-If you run several times `docker-compose up` and failing to load the DB for some reasons, remember to remove the `db_data` folder before trying again, as [the `init.sql` is only run once](https://stackoverflow.com/questions/53249276/docker-compose-mysql-init-sql-is-not-executed).
+## Tunning the environment
```
gs0510, line 40: Tuning*

pipeline/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+You can find more options for different configurations, posting message to slack, etc by running `--help` to the pipeline executable.
+
+
+### Troubleshooting
+
+On some computers, the docker containers have limited memory. If you get a docker error, try using a smaller data set for benchmarking, by changing the run arguments of the pipeline as follows:
+```ocaml
+Docker.pread ~run_args image
```
gs0510, line 39: This only worked for index benchmarks (since it had the `nb_entries` argument. I think we can scrap this section. Or rephrase it that they can modify the make bench target in the repo to run smaller dataset for benchmarks.

Docker.pread ~run_args image
  ~args:
    [
      "/usr/bin/setarch";
      "x86_64";
      "--addr-no-randomize";
      "_build/default/bench/bench.exe";
      "-d";
      "/dev/shm";
      "--json";
      "--nb-entries";
      "10000";
    ]

pipeline/bin/main.ml
```diff
@@ -12,7 +12,7 @@ module Source = struct
     in
     let github_token_path =
       Arg.required
-      @@ Arg.opt Arg.(some path) None
+      @@ Arg.opt Arg.(some file) None
```
gs0510, line 15: There's a slack path down a few lines which should also be a file, can you add that? Thanks! :)

pipeline/lib/pipeline.ml
```diff
@@ -84,22 +83,22 @@ let pipeline ~slack_path ~conninfo ?branch ?pull_number ~dockerfile ~tmpfs
       @ docker_cpuset_mems
     in
     let run_at = Ptime_clock.now () in
-    let+ output =
-      Docker.pread ~run_args image
-        ~args:
-          [
-            "/usr/bin/setarch"; "x86_64"; "--addr-no-randomize"; "make"; "bench";
```
gs0510, line 91: Should we investigate if we can disable ASLR inside docker containers as well?

gs0510 (commented):
Looks great! Thanks for this PR! Left a few tiny comments :)

pipeline/bin/main.ml
rizo: Done. I'm curious why isn't the slack path/URI passed as a string? Does it contain a secret?

pipeline/lib/pipeline.ml
rizo: I forgot to add this back. I was running into issues with this because I'm currently using an ARM machine.

Would you mind adding `setarch` and testing this on your machine?

rizo:
Thanks for the review, @gs0510 and @tatchi!

I think I've addressed all of the comments except https://github.com/ocurrent/current-bench/pull/55#discussion_r595056498

If one of you could try using `"/usr/bin/setarch"; "x86_64"; "--addr-no-randomize"; "make"; "bench";` for `Docker.pread` on an amd64 machine, I would appreciate that.

rizo:
@tatchi regarding your second point: https://github.com/ocurrent/current-bench/issues/64. Let's add this in a separate PR, if that's ok.

gs0510:
I tried `setarch` and it works! LGTM! :)

gs0510 (approved):

tatchi (approved):

PR #72, opened 2021-03-25
prevent legend from being hidden by the PR icon

tatchi:
## Before
Screenshot 2021-03-25 at 07 41 02

## After

Screenshot 2021-03-25 at 07 41 44


gs0510:
Looks good, thanks @tatchi!

PR #73, opened 2021-03-25
Responsive graphs

tatchi:
Current implementation uses a fixed graph size and tries to fill the current row with as many graphs as possible. When there's not enough space on the current row, we go to the next one. Problem is that we can end up with some space left not being used:

Screenshot 2021-03-25 at 07 52 34

This PR makes the graphs always taking up the entire row so that we avoid this "unused space". The [first commit](https://github.com/ocurrent/current-bench/commit/5d59ec090e3389470daedce606f876cb057d5b72) uses flexbox. I didn't find this solution ideal because the last row will always take the full width, meaning that graphs on this last row can have different sizes (if the number of graphs on this last row is different than the other rows)

Screenshot 2021-03-25 at 07 15 51

[Second commit](https://github.com/ocurrent/current-bench/commit/2c7d96fbb67dd9e10bd3c986aa7a2a29116aa03c) uses CSS grid, that makes every graph have the same size, even for the last row:

Screenshot 2021-03-25 at 08 14 35

A couple of other screenshots taken at different screen resolutions:

Screenshot 2021-03-25 at 08 11 29
Screenshot 2021-03-25 at 08 10 20
Screenshot 2021-03-25 at 08 10 38
Screenshot 2021-03-25 at 08 10 54

### DIFF
--- a/frontend/src/BenchmarkTest.res
+++ b/frontend/src/BenchmarkTest.res
@@ -235,7 +235,10 @@ let make = (
         delta == 0.0 ? "Same as master" : deltaToString(delta) ++ " vs master"
       )
 
-      <div key=metricName>
+      <div
+        key=metricName
+        // one: [Css.flex3(~grow=1.0, ~shrink=1.0, ~basis=#percent(0.0))],
+        className={Sx.make([Sx.unsafe("minWidth", "400px"), Sx.unsafe("flex", "1")])}>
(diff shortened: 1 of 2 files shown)


gs0510:
CSS grid definitely makes them look better. Thanks @tatchi!

gs0510 (approved):

PR #74, opened 2021-03-30
Use error bars in dygraphs.

rizo:
WIP implementation of error bars for metrics with multiple values. The graph shows the mean; the lower bound is the minimum, the upper bound is the maximum.

See this example in dygraphs: https://dygraphs.com/gallery/#g/temperature-sf-ny

This is how it looks in our dashboard:

Screenshot 2021-04-27 at 17 18 36

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -69,7 +69,7 @@ summary::-webkit-details-marker {
   /* font-size: 14px; */
   z-index: 10;
   line-height: normal;
-  overflow: hidden;
+  overflow: visible;
   color: black;  /* replaces old axisLabelColor option */
 }
 
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -226,7 +226,7 @@ module RepoView = {
               <Welcome />
             </Column>
           </>
-        | Some(repoId) when !(repoIds->BeltHelpers.Array.contains(repoId)) =>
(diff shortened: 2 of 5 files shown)


rizo:
This PR also makes minor changes to the graph rendering. This is because with error-bar we cannot use filled graphs. I took the opportunity to simplify things a bit.

In particular, the graphs now always show the origin, fixing https://github.com/ocurrent/current-bench/issues/103.

**Before:**
<table>
<tr>
<td>Screenshot 2021-04-27 at 18 21 18
</td>

<td>
Screenshot 2021-04-27 at 18 23 03
</td>

<td>
Screenshot 2021-04-27 at 18 23 31
</td>

</tr>
</table>

**After:**

<table>
<tr>
<td>
Screenshot 2021-04-27 at 18 21 13
</td>

<td>
Screenshot 2021-04-27 at 18 24 30
</td>

<td>
Screenshot 2021-04-27 at 18 23 35
</td>

</tr>
</table>


This makes the small variations barely visible on some graphs as you can see. But I think it's a reasonable trade-off.

rizo:
Related to the multi-value datapoints: it would be good to show min/avg/max in the tooltip. I'll add this in a follow-up PR.

ctk21:
This is great, as in #103, I like seeing the origin. 
For example in the dashboard example, you immediately see that the big (relative) change is in `merge_duration_us`.

gs0510 (approved):

PR #75, opened 2021-03-30
Fix production deployment; add instructions.

rizo:


gs0510 (approved):

rizo:
(Please do not merge this yet, still testing this out.)

HACKING.md
```diff
@@ -146,3 +146,57 @@ pipeline_1    |                             "--" "/tmp/git-checkout3219a5b2"
 ```
 
 > **WARNING**: When committing changes to the local test repo, make sure you are doing so in the main top-level git repository.
+
```
gs0510, line 149: Something missing from this doc is that one can also test the app on their (single) github repo given that the user has a personal github token to the repo they are testing for. Perhaps we should add that also.

rizo: That's true.

I tried to focus on the two main environments we would use during development and production. Let's add these instructions in the future.

PR #82, opened 2021-03-31
update to `@urql/rescript` and `@rescript/react`

tatchi:
`reason-urql` is now compatible with `urql@2` and [has been renamed to `@urql/rescript`](https://github.com/FormidableLabs/rescript-urql/releases). It now supports the new `@rescript/react` package.

I also updated a few other dependencies such as `vite` and `bs-css`.

### DIFF
--- a/frontend/bsconfig.json
+++ b/frontend/bsconfig.json
@@ -18,15 +18,14 @@
   "suffix": ".bs.js",
   "namespace": true,
   "bs-dependencies": [
-    "reason-react",
+    "@rescript/react",
     "bs-css",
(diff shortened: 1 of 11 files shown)


tatchi:
I have the impression that when switching from one PR to another one, the data takes a bit longer to show up. Especially the PR's number in the breadcrumb which doesn't change instantaneously...

gs0510 (approved):

gs0510:
Thanks @tatchi! Do you suspect the lag for data showing up is because of this dependency?

Issue #88, opened 2021-04-02
Fails to build development environment


d4hines:
Hello,

current-bench fails to start for me.

I'm on Windows 10, running Docker version 20.10.5.

The contents of ./environments/development.env are:
```
OCAML_BENCH_DOCKER_CPU=1
OCAML_BENCH_DB_PASSWORD=docker
OCAML_BENCH_GRAPHQL_KEY=secret
OCAML_BENCH_TARGET_ARCH=amd64
```

I run `make start-development` and get the following output (errors way down at bottom):
```
PS C:\Users\d4hin\repos\current-bench> make start-development
docker-compose \
        --project-name="current-bench" \
        --file=./environments/development.docker-compose.yaml \
        --env-file=./development.env \
        up \
        --remove-orphans \
        --build
Building db
failed to get console mode for stdout: The handle is invalid.
[+] Building 11.2s (7/7) FINISHED
 => [internal] load build definition from Dockerfile                       0.0s 
 => => transferring dockerfile: 151B                                       0.0s 
 => [internal] load .dockerignore                                          0.0s 
 => => transferring context: 2B                                            0.0s 
 => [internal] load metadata for docker.io/library/postgres:12             1.7s 
 => [internal] load build context                                          0.0s 
 => => transferring context: 87B                                           0.0s 
 => [1/2] FROM docker.io/library/postgres:12@sha256:2f3f78532c9cc5435d1cf  9.2s 
 => => resolve docker.io/library/postgres:12@sha256:2f3f78532c9cc5435d1cf  0.0s 
 => => sha256:2f3f78532c9cc5435d1cf9c8f5fb1409f9abd43e5b7 1.86kB / 1.86kB  0.0s
 => => sha256:75646c2fb4101d306585c9b106be1dfa7d82720ba 27.14MB / 27.14MB  1.0s
 => => sha256:7e98825f6d67159fc035c0db985619c0635cc7edb89 1.80kB / 1.80kB  0.3s
 => => sha256:6ef141d922bc14123d24ddec34aa11d0f360702a2 10.63kB / 10.63kB  0.0s
 => => sha256:4217f50ff2ca28b7c3dddb25372379c97093df532c7 3.25kB / 3.25kB  0.0s
 => => sha256:2355d0ffeb5531909b8910928bcbaa874e9464657c7 4.18MB / 4.18MB  0.4s
[...]

PR #91, opened 2021-04-07
Add health check to ensure postgres db is running

gs0510:
Addresses #88


rizo (approved):
Looks good!

One thing I would change is the healthcheck interval:

- `interval: 10s` seems too much for our use case. According to the [docs](https://docs.docker.com/engine/reference/builder/#healthcheck) `interval` is used to wait before even attempting any checks (and after every attempt). I think 5s is probably better here as PostgreSQL should start pretty quickly.

Issue #64, opened 2021-03-16
Automatically detect the backend URI


rizo:
@tatchi :
> When I started the frontend, I was a bit surprised to actually see data from http://autumn.ocamllabs.io. That would be nice to grab the backend URL from env (it's currently hardcoded in the code). Vite allows reading .env files but from what I understand, they should be located at the root of the project (within frontend/ in our case). Maybe we could copy (and rename) the env file e.g development.env within the frontend folder during one of the docker steps? Not sure though if that's a good idea 😅

---

Another (very simplistic) option is to decide the backend URI based on the the current window location in the app. If it's `localhost`, use `localhost`, if it's not, use the production URI.

gs0510:
We now use the URI from env variables; do we still want it to be detected using current window location? @rizo

PR #94, opened 2021-04-08
use graphql url from env

tatchi:
Addresses #64 

Vite can [load variables](https://vitejs.dev/guide/env-and-mode.html#env-files) from an `.env` file defined at the root of the project (`/frontend` in our case).

This PR allows defining a variable `VITE_GRAPHQL_URL` within the `environments/development.env` or `environments/development.env` file. During the `docker-compose` workflow, an `.env` file that contains this variable is created in the `frontend` folder, so that `vite` can use it when bundling the app.

That's nice because we can define all the variables in one file for the `docker-compose` workflow. One drawback though is that outside of the `docker-compose` workflow, we would still have to manually add an `.env` file under the `/frontend` folder.

The other solution would be to use this manually created `.env` file for both scenarios. We would lose the benefit of having only one file where we define env variables for the `docker-compose` workflow, but that might be simpler...

What do you think?


frontend/Dockerfile
```diff
@@ -10,6 +12,8 @@ RUN yarn install
 # add app
 COPY . /app/
 
+RUN echo "VITE_GRAPHQL_URL=${VITE_GRAPHQL_URL}" > /app/.env
```
tatchi, line 15: Not sure if that's the best way of doing that... I first tried to define a `volumes` in the `docker-compose.yaml` file and somehow trying to use it here but I wasn't able to make it work.

environments/development.docker-compose.yaml
```diff
@@ -21,7 +21,10 @@ services:
       HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
       HASURA_GRAPHQL_ADMIN_SECRET: ${OCAML_BENCH_GRAPHQL_KEY?required}
   frontend:
-    build: ../frontend
+    build: 
+      context: ../frontend
+      args:
+        VITE_GRAPHQL_URL: ${VITE_GRAPHQL_URL}
```
gs0510, line 27: We should add the required tag here as well.

frontend/Dockerfile
gs0510: Since we'll need this in the root, I think it's not a bad idea.

rizo:
@tatchi would it be useful to set the env var to `localhost` by default?

tatchi:
> @tatchi would it be useful to set the env var to `localhost` by default?

Yes definitely, will create a PR :)

PR #95, opened 2021-04-08
set up public role + metadata

tatchi:
This PR sets up a `public` role via the `HASURA_GRAPHQL_UNAUTHORIZED_ROLE=public` env variable. More info [here](https://hasura.io/docs/latest/graphql/core/auth/authentication/unauthenticated-access.html#no-auth-setup)

Problem is that we still need to add this `public` role on the Hasura interface:

image

Turns out that we can actually export `metadata` from a current Hasura instance and re-apply them during the `docker-compose` workflow, so that we don't need anymore to manually track the `benchmark` table nor add the `public` role.

First step was to init the `hasura-server` project with the `hasura-cli`:

```bash
# https://hasura.io/docs/latest/graphql/core/migrations/migrations-setup.html#step-1-install-the-hasura-cli
hasura init hasura-server --endpoint http://localhost:8080/
```

Then export the metadata:

```bash
hasura metadata export --endpoint http://localhost:8080/ --admin-secret "my_secret"
```

Then use a specific hasura docker image `hasura/graphql-engine:v1.3.3.cli-migrations-v2` and mount a `hasura-metadata` folder from the locally exported `metadata` folder. More info [here](https://hasura.io/docs/latest/graphql/core/migrations/config-v2/advanced/auto-apply-migrations.html#auto-apply-migrations-metadata-when-the-server-starts-config-v2)


tatchi:
Note that I didn't make any changes to the `http://autumn.ocamllabs.io/` instance. I can add the `public` role if needed via the interface but the `HASURA_GRAPHQL_UNAUTHORIZED_ROLE` stuff would probably need a restart?

gs0510:
Setting this up on autumn now, will report back and merge :crossed_fingers:

d4hines:
I like this a lot better. As I've never worked with Hasura before, it took quite a while to figure out that I had to manually track them in first place! Good work

PR #111, opened 2021-04-27
Install only bench dependencies for dune

gs0510:
A hack to get dune benchmarks to run until we support custom dockerfiles. The dune benchmarks currently fail when trying to install dune dependencies (which we don't need).

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -39,9 +39,10 @@ let read_channel_uri p =
 (* Generate a Dockerfile for building all the opam packages in the build context. *)
 let dockerfile ~base ~repository =
   let opam_dependencies =
-    if String.compare repository "dune" == 0 then
-      format_of_string "opam install ./dune-bench.opam -y --deps-only  -t"
-    else format_of_string "opam install -y --deps-only -t ."
+    (* FIXME: This should be supported by a custom Dockerfiles. *)
+    if String.equal repository "dune" then
+      "opam install ./dune-bench.opam -y --deps-only  -t"
+    else "opam install -y --deps-only -t ."
   in
   let open Dockerfile in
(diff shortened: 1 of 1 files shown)


pipeline/lib/pipeline.ml
```diff
@@ -37,7 +37,12 @@ let read_channel_uri p =
   Util.read_fpath p |> String.trim |> Uri.of_string |> Current_slack.channel
 
 (* Generate a Dockerfile for building all the opam packages in the build context. *)
-let dockerfile ~base =
+let dockerfile ~base ~repository =
+  let opam_dependencies =
+    if String.compare repository "dune" == 0 then
```
rizo, line 42: I think this deserves a comment – either a `HACK` or `FIXME` to make it clear that this is not ok long-term.

rizo: Also, why not `String.equal`?

rizo: I feel like having this as format is strange.

Maybe let's use a string here and move the conversion down as:

```ocaml
  @@ run "%s" opam_dependencies
```

rizo:
@gs0510 this is an ok temporary solution.

Because this is a temporary solution I suggest:

1. we create a high-priority issue to track the fix;
2. we describe the possible solutions.

My preferred solution at the moment is supporting custom dockerfiles per repository, i.e., checking if `bench.docker` exists at the root of the project and using that to build the image.

There are probably other solutions, let's list them here.

@MagnusS I think you had some suggestions/reservations on this, if I recall correctly.

MagnusS:
@rizo my main concern iirc was just that if what we really want is a specific opam switch or standard compiler version, it may be better to support specifying that directly (we may also want to support a matrix of versions) -- but probably better to discuss on an issue, PRs tend to get lost :)

PR #114, opened 2021-04-30
Check default ref for running benchmarks

gs0510:


pipeline/lib/pipeline.ml
```diff
@@ -188,15 +188,17 @@ let process_pipeline ~(docker_config : Docker_config.t) ~conninfo
         |> let> api, repo = repo in
            Github.Api.refs api repo
       in
+      let default_branch = Github.Api.default_ref refs in
```
rizo, line 191: I'm a little bit confused by this. Where is `default_branch` actually used?

I think the code below should be:

```
match key with
| `Ref branch where branch = default_branch -> pipeline ~head ~branch ()
```

pipeline/lib/pipeline.ml
```diff
@@ -234,11 +238,11 @@ let process_pipeline ~(docker_config : Docker_config.t) ~conninfo
               (fun key head _ ->
                 let head = `Github head in
                 match key with
-                | `Ref "refs/heads/master" -> pipeline ~head ~branch:"master" ()
+                | `Ref default_branch -> pipeline ~head ~branch:default_branch ()
```
rizo, line 241: Ditto.

rizo (changes requested):
I think there's a small mistake with how the default branch is matched. Once fixed, this should be good to merge.

pipeline/lib/pipeline.ml
gs0510: Yep you're right, I got confused while reasoning about default_refs

pipeline/lib/pipeline.ml
```diff
@@ -193,15 +193,16 @@ let process_pipeline ~(docker_config : Docker_config.t) ~conninfo
...
            Github.Api.refs api repo
       in
+      let default_branch = Github.Api.default_ref refs in
+      let ref_map = Github.Api.all_refs refs in
       Github.Api.Ref_map.fold
         (fun key head _ ->
           let head = `Github head in
           match key with
-          | `Ref "refs/heads/master" -> pipeline ~head ~branch:"master" ()
+          | `Ref branch -> if branch = default_branch then pipeline ~head ~branch () else Current.return ()
```
rizo, line 202: This will capture `refs/heads/master` as `branch` instead of using `"master"` as in the current version.

I think we should keep the branch name the same, i.e., continue using `"master"`.

Ideally this should apply to the default branch too.

rizo (commented):
Nothing major, but I highlighted a breaking change in the way the branch name is captured. Considering that the branch name is used in the UI we should use short version of the ref.

PR #118, opened 2021-05-07
Add Github actions to current-bench.

gs0510:
Basic template to build pipeline for now, will add commands to build rescript as well.


gs0510:
Closing because we'll use ocaml-ci instead! I'll try to figure out what to do for rescript in the meantime.

PR #122, opened 2021-05-12
fix conf-zlib dependency issue

shubhamkumar13:
This PR fixes the build error which occurred while trying to setup `current-bench` using the postgres docker container.

The PR adds `zlib1g-dev` which is required by `conf-zlib`

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -42,7 +42,7 @@ let dockerfile ~base =
   from (Docker.Image.hash base)
   @@ run
        "sudo apt-get update && sudo apt-get install -qq -yy libffi-dev \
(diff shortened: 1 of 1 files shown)


rizo:
Hey! Thanks for the PR!

Could you elaborate on what the error was and in what circumstances it was occurring?

This patch adds `zlib1g-dev` to the base docker image we use to run the benchmarks. I don't understand the relation to Postgresql.

Could you help me reproduce the error?

rizo:
Clarified in a dev meeting: this will not be merged. There will be a solution with custom docker files that solves this.

rizo:
https://github.com/ocurrent/current-bench/issues/11

shubhamkumar13:
Closing this. Thanks for the review @rizo

PR #123, opened 2021-05-17
Add symlinks for ocaml-ci to run.

gs0510:


gs0510:
ocaml-ci is failing on alpine for everyone because the capnp-proto package is missing. 

The other failure is a linter failure, because refmt is missing. refmt is an npm package so can't be picked up by ocaml-ci, one solution I can think of is packaging  refmt as an opam package or getting `dune @fmt` to only format the ocaml files and not the reason files. (I don't know if either of these are doable, I'd need to look into it more!)

rizo (approved):

PR #133, opened 2021-05-21
Attach a volume to manage data dependencies.

gs0510:
I'm not really sure if the dev version needs a volume, maybe we can just have this for the prod env?

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -187,6 +187,20 @@ Assuming that the private key is saved in `ocaml-bench-github-key` at the projec
 $ docker run -it --rm -v $PWD/ocaml-bench-github-key:/mnt/ocaml-bench-github-key -v current-bench_pipeline_var:/mnt/pipeline_var alpine cp /mnt/ocaml-bench-github-key /mnt/pipeline_var/ocaml-bench-github-key
 ```
 
+#### Adding a docker volume for data dependencies
+`current-bench` supports benchmarks that have data dependency. You can add the dependencies to a
+docker volume which is then mounted to the pipeline container in `docker-compose`.
+
+To create a docker volumes you should run:
+```
+$ docker volume create bench-data
+```
+You can add the data required to your `bench-data` volume by inspecting the volume and copying the data directly to the path listed in the inspect results.
+```
+$ docker volume inspect
+```
+
+The underlying assumption here is that the `make bench` target would know that the dependencies live in the `bench-data` folder inside the container.
 
 ### Starting the production environment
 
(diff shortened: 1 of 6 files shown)


HACKING.md
```diff
@@ -187,6 +187,20 @@ Assuming that the private key is saved in `ocaml-bench-github-key` at the projec
 $ docker run -it --rm -v $PWD/ocaml-bench-github-key:/mnt/ocaml-bench-github-key -v current-bench_pipeline_var:/mnt/pipeline_var alpine cp /mnt/ocaml-bench-github-key /mnt/pipeline_var/ocaml-bench-github-key
 ```
 
+#### Adding a docker volume for data dependencies
+`current-bench` supports benchmarks that have data dependency. You can add the dependencies to a
+docker volume which is then mounted to the pipeline container in `docker-compose`.
+
+To create a docker volumes you should run:
+```
+$ docker volume create bench-data
```
rizo, line 196: Minor point, but let's suggest the volume to be called `current-bench-data`.

rizo (changes requested):
I think this approach is lacking data namespacing between different projects.

With the current implementation, the loaded data will be shared by all projects. This is worrying because:

1. we should ensure data is private;
2. projects can override/delete data loaded by other projects;
3. it  is unclear what project requires what data.

I understand this feature is still in early stages, but I think we should come up with a convention that provides a baseline with some kind of namespacing.

For example, we could require that projects work with data only in directories that match the following pattern: `/mnt/current-bench-data/$owner_name/$repo_name`.

I think we should also think how this will transition into a more elaborate system in the future.

gs0510:
@rizo I think something I can also do is to explore how can we do this without using a docker volume but using an ocurrent step in the pipeline? Does that sound good? I think with the docker volume it will be a bit precarious because we'd need to host the data dependencies on our machines and ensure that they work with the benchmarks.

gs0510:
Screenshot which shows the data in <owner>/<repo_id> when mounted on the pipeline docker container.

screenshot

rizo:
> @rizo I think something I can also do is to explore how can we do this without using a docker volume but using an ocurrent step in the pipeline? Does that sound good? I think with the docker volume it will be a bit precarious because we'd need to host the data dependencies on our machines and ensure that they work with the benchmarks.

I think this makes sense. In essence we want to have multiple stages in the pipeline each one producing any artefacts necessary with potential dependencies in between. Namely:

* Build: the opam file and the source code of the project produces a docker image.
* Dependencies: some kind of dependency specification file/format (with hashes) describes where to download the dependencies from.
* Run: the built docker image and the dependencies are provided to start running the benchmark producing the benchmark results JSON.

An important aspect is that we need an ability to fetch dependencies on demand and cache them. I think ideally we should clean dependencies periodically and allow the pipeline to refetch them when needed.

In any case, let's merge this PR. I think it's a good starting point.

PR #134, opened 2021-05-30
set subtitle min-height

tatchi:
This PR adds a `min-height` to the subtitle so that even if it's empty, space is reserved and graphs keep the same height.

### Before
Screenshot 2021-05-30 at 07 34 31

### After 

image


gs0510:
Awesome, thanks @tatchi !

PR #137, opened 2021-06-06
docker-compose: expose should be an array

tatchi:
* First commit fixes this error:

Screenshot 2021-06-06 at 09 59 06

* Second commit remove a duplicated value
* Third commit format the yaml files. I can remove this commit if you prefer not to have these files formatted by prettier


gs0510:
Thanks @tatchi! I'd been meaning to fix it 😁

PR #138, opened 2021-06-06
add stop-development command

tatchi:
Similar to the `stop-production`, this PR adds a `stop-development` command.


gs0510:
Thank you!!

Issue #62, opened 2021-03-15
When multiple benchmarks are produced ensure that there are no name collisions


rizo:
By default `benchmark_name` is set to `NULL`. We do this even if there are multiple objects produced by `make bench`. This most certainly is a mistake and we should require the output to have the `benchmark_name` field with multiple benchmark results.

PR #153, opened 2021-10-07
manage name collisions in the current-bench pipeline

shubhamkumar13:
This PR aims to resolve https://github.com/ocurrent/current-bench/issues/62, it errors out when there is a name collision in the toplevel name field of the benchmark json.

https://github.com/ocurrent/current-bench/issues/62 also states that the benchmark name is a NULL type in the database schema - https://github.com/ocurrent/current-bench/blob/12b86dc73edaa94a81ace014b7b845b2cc4dc9c2/pipeline/db/migrations/20210013150054_add_benchmarks_table.up.sql#L7

Which means it is currently optional and can be left out as NULL in the database table and the field is omitted in the json produced in the `pread_log` of the current-bench UI.

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -9,11 +9,15 @@ module Benchmark = Models.Benchmark
 
 let ( >>| ) x f = Current.map f x
 
-let is_duplicate elem lst =
-  lst
-  |> List.map (fun x -> if x = elem then 1 else 0)
-  |> List.fold_left (fun acc x -> acc + x) 0
-  |> fun x -> x > 1
+let is_duplicate (elem : string option) (tbl : (string option, int) Hashtbl.t) =
+  Hashtbl.find_all tbl elem |> List.length |> fun x ->
+  if x > 1 then
+    Result.error
+      (Printf.sprintf
+         "Benchmark %s has a duplicate name, please change the name of the \
+          benchmark to something unique"
+         (match elem with Some x -> x | None -> "None"))
+  else Result.ok ()
 
 module Source = struct
   type github = {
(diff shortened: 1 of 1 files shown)


pipeline/lib/pipeline.ml
```diff
@@ -9,6 +9,9 @@ module Benchmark = Models.Benchmark
 
 let ( >>| ) x f = Current.map f x
 
+let is_duplicate elem lst = lst |>
```
gs0510, line 12: Instead of checking this in the list every time, can we just have a hashtable and check if the value exists there? It'll be faster and there won't be a lot of names so not a memory problem either.

art-w:
Thanks, this validation of the json was much needed! I think the code could be a lot simpler if you extracted the logic into its own function :) (rather than interleaved with the existing code)

pipeline/lib/pipeline.ml
```diff
@@ -116,23 +126,37 @@ let pipeline ~slack_path ~conninfo ?branch ?pull_number ~dockerfile ~tmpfs
...
                |> Yojson.Safe.Util.to_string_option
              in
-             Yojson.Safe.Util.(member "results" output_json)
-             |> Yojson.Safe.Util.to_list
-             |> List.map
-                  (Benchmark.make ~duration ~run_at ~repo_id ~benchmark_name
-                     ~commit ?pull_number ?build_job_id ?run_job_id ?branch)
-             |> List.iter (Models.Benchmark.Db.insert db));
+             match is_duplicate benchmark_name benchmark_tbl with
+             | Error e -> raise @@ Failure e
```
art-w, line 152: The `raise` can be done while building the Hashtbl (we can detect that there is a duplicate before inserting)

shubhamkumar13:
Updated the `is_duplicate` function to `validate_json` where `raise` happens while adding benchmarks to the `Hashtbl`.

Also created 2 new functions `get_benchmark_name` and `get_result_list` : https://github.com/ocurrent/current-bench/blob/ef412efa45210608b383babfc684b6fc8e2ac924/pipeline/lib/pipeline.ml#L12-L16
which was previously interleaved in the existing code :
https://github.com/ocurrent/current-bench/blob/40d871762dfe23d554258d842c20d97ccf511c5b/pipeline/lib/pipeline.ml#L125-L135

This also increase readability in the `validate_json` function :
https://github.com/ocurrent/current-bench/blob/ef412efa45210608b383babfc684b6fc8e2ac924/pipeline/lib/pipeline.ml#L18-L30

Thanks for the feedback @art-w, let me know if this seems ok :)

PR #155, opened 2021-10-12
Clarify development.env.template to make local development easier

punchagan:
This PR has a couple of commits to make it easier to get the local development environment started locally.

PR #158, opened 2021-10-13
Improve the local development environment for front-end development

punchagan:
This PR allows running the bbs watcher and the node (vite) development server inside the frontend docker container, instead of running anything locally. The commit messages describe the changes in more detail. 

It also reverts the changes to the development.env.template in PR #155 to expose the port 8082 where the vite (node) server runs.


gs0510:
> something to do with the value set for OCAML_BENCH_GRAPHQL_URL in your development.env
Yes it did! My bad, thanks!

> What other modes would be useful here? Am I missing something here?
I'm not sure, but perhaps just having a non-watch mode also, we don't have to do that now, we can cross that bridge when we get to it 😁

punchagan:
> I'm not sure, but perhaps just having a non-watch mode also, we don't have to do that now, we can cross that bridge when we get to it

Cool! I've never actually needed this in the past, and crossing the bridge if and when we get to it sounds good! :+1: 

Thanks for merging this PR. :)

Issue #105, opened 2021-04-26
Abstract out Docker image name from pipeline/lib/pipeline.ml to environments


shakthimaan:
The "ocaml/opam" Docker image name is hard-coded in `pipeline/lib/pipeline.ml`. Kindly abstract it out to an environment variable, so that Multicore OCaml can specify `ocaml/opam:ubuntu-20.10-ocaml-4.10`.
```
  let dockerfile =
    let+ base = Docker.pull ~schedule:weekly "ocaml/opam" in
    `Contents (dockerfile ~base)
  in
```

gs0510:
Should this be done in the Sandmark pipeline as well (along with the support for `--privileged` flag) since all projects we have running now run with the `ocaml/opam` image.

PR #164, opened 2021-10-18
Custom dockerfiles

art-w:
This PR adds support for custom dockerfiles: If a `Dockerfile` is present at the root of the project, then it will be used -- otherwise we keep the existing setup based on `ocaml/opam`.

This would allow to run benchmarks on ocaml multicore https://github.com/ocurrent/current-bench/issues/105 and to install a different set of dependencies https://github.com/ocurrent/current-bench/issues/11


pipeline/lib/custom_dockerfile.ml
```diff
@@ -0,0 +1,88 @@
...
+  @@ run "eval $(opam env)"
+
+let weekly = Current_cache.Schedule.v ~valid_for:(Duration.of_day 7) ()
+
+let dockerfile ~repository =
+  let+ base = Docker.pull ~schedule:weekly "ocaml/opam" in
+  `Contents (dockerfile ~base ~repository)
+
+let dockerfile ~src ~repository =
+  let custom_dockerfile = Fpath.v "Dockerfile" in
```
gs0510, line 85: We should perhaps look for `bench.Dockerfile` rather than `Dockerfile` :)

pipeline/lib/custom_dockerfile.ml
```diff
@@ -0,0 +1,88 @@
...
+
+let file_exists ~src filename =
+  Current.component "file-exists %S" (Fpath.to_string filename)
+  |> let> commit = src in
+     Cache.get () { File_exists.Key.commit; filename }
+
+(* Generate a Dockerfile for building all the opam packages in the build context. *)
+let dockerfile ~base ~repository =
+  let opam_dependencies =
+    (* FIXME: This should be supported by a custom Dockerfiles. *)
```
shubhamkumar13, line 60: This comment is not needed anymore

gs0510: It is until we add a dockerfile to dune :) I'll open an issue to remove it once we do,

gs0510:
Thanks for the PR @art-w! Can we also add to the documentation that the users can have custom Dockerfiles now (Along with an example of what the dockerfile should look like). Thank you!

PR #166, opened 2021-10-19
Add eval command so make can find dune

gs0510:


art-w:
It's so weird that it worked before, but good find!

PR #168, opened 2021-10-20
Add a script to be able to run docker-compose command easily

punchagan:
The project setup uses files in the environment directory, which are
non-standard, and running docker-compose commands requires passing 3 arguments
to get the command to run correctly.  This commit adds a script which makes it
easy to run docker-compose commands. This could alternatively be achieved by
adding a "run" target to the Makefile, but it seems clunkier than adding a
separate script.

This would also make it easy to run variations of the start-development and
stop-development commands.

The script can be used to run commands, like

   ./dc.sh down -v
   ./dc.sh exec db psql -U docker

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -57,11 +57,11 @@ Start the docker-compose environment:
 ```
 $ make start-development
 docker-compose \
-		--project-name="current-bench" \
-		--file=./environments/development.docker-compose.yaml \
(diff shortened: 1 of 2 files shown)


gs0510:
Looks good! Thanks!

gs0510 (approved):

PR #170, opened 2021-10-20
Fix broken down migration for add_duration_column_to_benchmarks

punchagan:
The migration was added in ad094ce86751720a2c976150d1096fc4b8863869 but the
down migration was an empty file. This causes `omigrate down` to fail. The
current commit fixes this down migration to drop the duration column being
added in the up migration.


gs0510 (approved):
good catch @punchagan! Thanks for the PR!

PR #171, opened 2021-10-20
Use 4.13 and install graphviz in pipeline docker

gs0510:


pipeline/Dockerfile
```diff
@@ -9,14 +9,15 @@ RUN sudo apt-get update && \
...
-    capnproto \
     libsqlite3-dev \
+    capnproto \
     libcapnp-dev
 
 WORKDIR /mnt/project
 
 # Build dependencies.
 COPY --chown=opam:opam pipeline.opam pipeline.opam
+RUN sudo mv /usr/bin/opam-2.1 /usr/bin/opam
```
gs0510, line 20: Also using 2.1 opam, doesn't really make a difference since we are not using any 2.1 features, yet

PR #176, opened 2021-10-21
Fix: pull "ocaml/opam" only once

art-w:
The graphviz visualization suggests that we are pulling the "ocaml/opam" image for each pipeline, rather than once every week. It's probably cached so this should'nt make a big difference beyond reducing the size of the graphviz automata.


gs0510 (approved):
Awesome! Thanks!

PR #177, opened 2021-10-21
Collapse pipeline nodes

art-w:
The graphviz automata is a bit large in production, so maybe we should collapse `[+]` the pipelines to reduce the noise :)


gs0510:
Nice, I didn't know this existed 😁 Thanks! :)

PR #188, opened 2021-10-29
Optimize release: do not rerun all benchmarks

art-w:
When we deploy a new release and the pipeline has changed, ocurrent reruns everything since it could potentially produce a different result. This is not exactly desirable since it takes a very long time and most of our pipeline changes are not impactful on past PR.

This patch checks if we already have some benchmarks saved in sql for this repo/commit... and if so, it skips the rerun!

Also included is a small patch to `opam pin` since it may trigger some huge side effects (and it's a typical example of a pipeline change where we do not want to rerun everything since it's just a bugfix.)

### DIFF
--- a/pipeline/lib/custom_dockerfile.ml
+++ b/pipeline/lib/custom_dockerfile.ml
@@ -112,7 +112,7 @@ let dockerfile ~base ~dependencies ~opam_file =
   base_dockerfile ~base
   @@ install_static_dependencies
   @@ add_workdir
-  @@ run "opam exec -- opam pin ."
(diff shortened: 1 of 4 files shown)


punchagan (approved):
Looks good to me!

punchagan:
> I guess the trade-off here is either making a bunch of queries to the db vs re-running the benchmarks and the queries to the db seem less expensive :)

We could probably add an index here to make these queries even less expensive. Something like

```sql
CREATE INDEX benchmarks_repo_id_commit_idx ON benchmarks (repo_id, commit);
```

PR #189, opened 2021-10-29
Add current_bench_json.ml to share json validation between backend/frontend

art-w:
Currently the json format is defined partially in the pipeline (which doesn't care about validating the metrics, just the general shape) and the frontend (which only cares about the metrics). We've had situations where the pipeline would succeed even though the metrics were unreadable by the frontend.

I tried to setup a way for the backend and frontend to agree on what the entire json should be, such that the pipeline can indicate a problem if the user submits bad metrics, and that both can stay uptodate with the latest schema. The result is in `pipeline/lib/current_bench_json.ml` and by the magic of symbolic link, it's also available in the Rescript frontend (I had to fight both buildsystem and docker, so I hope it works now!)

There are some strong limitations in this shared file. It can not depend on any other file and Rescript only understands OCaml 4.06 with a limited stdlib (no regex?!.. and no Yojson or other dependencies)

Perhaps it can also be useful if at some point we expose a user library to help projects produce the latest version of the json.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+frontend/node_modules
--- a/environments/development.docker-compose.yaml
+++ b/environments/development.docker-compose.yaml
@@ -35,7 +35,8 @@ services:
     cpuset: "${OCAML_BENCH_PLATFORM_CPUSET}"
   frontend:
     build:
-      context: ../frontend
+      context: ..
+      dockerfile: ./frontend/Dockerfile
       target: dev
     environment:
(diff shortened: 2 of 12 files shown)


punchagan:
> An alternative solution for the frontend would be to run a database migration on previous entries to upgrade them to the latest json format -- and then we would only have to maintain the latest decoder in the frontend.

We would also need to make sure that each time we update the version schema, we "re-run" this migration to migrate the previous latest version in the DB to the current latest version.

gs0510:
This is also okay for me to be merged after rebase, I'll add an issue for migrating v0 to v1 so we can discard the v0 decoder :) Thanks @art-w!

gs0510 (approved):

PR #190, opened 2021-10-29
Allow selected projects to run on more than one CPU

art-w:
This patch adds support for running some projects on more than one core... until we find a better way to configure this setting!

The following commands can be helpful to determine which value to give to `OCAML_BENCH_DOCKER_CPU` (where the benchmark will run):
```
$ cat /sys/devices/system/cpu/isolated
2-3

$ cat /sys/devices/system/cpu/present   
0-7
# ^^^ all the cpus available, subtract isolated to define OCAML_BENCH_PLATFORM_CPUSET
```


pipeline/lib/pipeline.ml
```diff
@@ -150,8 +154,13 @@ let docker_make_bench ~run_args ~repository image =
         "opam exec -- make bench";
       ]
 
-let pipeline ~conninfo ~run_args ~repository =
+let is_multicore repository =
```
gs0510, line 157: Can we also pass the multicore repository as an optional argument? :)

gs0510 (changes requested):

pipeline/lib/pipeline.ml
```diff
@@ -150,8 +154,13 @@ let docker_make_bench ~run_args ~repository image =
...
+  List.mem (Repository.info repository) [ "art-w/ocaml"; "local/local" ]
```
gs0510, line 158: Could we do something that is more parametrized here? :D Perhaps passing the repositories where you can have more than one CPU as a list? :)

gs0510 (approved):
LGTM, thank you!

gs0510:
Once you rebase this @art-w, it can be merged :)

PR #191, opened 2021-11-01
Simplify production deployments to be a single command

punchagan:
This PR adds a make target to redeploy a production deployment. 

1. It builds the containers required, before stopping the currently running instance, to reduce the down-time of the service. 

2. It adds a commit to use `omigrate` to run the migrations instead of manually running the migrations on the Postgres DB.  The production DB may need to be updated to mark the existing migrations as already run. I'm not sure why we used the manual approach so far, and this seems better, unless I'm missing something. I don't have access to the production servers, right now, to check what needs to be done there to make this PR usable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,14 @@ start-production:
 		--detach \
 		--build
 
+.PHONY: build-production
+build-production:
+	docker-compose \
+		--project-name="current-bench" \
+		--file=./environments/production.docker-compose.yaml \
+		--env-file=./environments/production.env \
+		build
+
 .PHONY: stop-production
 stop-production:
 	docker-compose \
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,12 @@ stop-production:
 ./local-test-repo/.git:
 	cd ./local-test-repo/ && git init && git add . && git commit -m "Initial commit."
 
+.PHONY: run-migrations
+run-migrations:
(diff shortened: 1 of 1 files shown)


gs0510:
@punchagan I think perhaps we can have two targets, one redeploy with migration and one without, we don't need to the run the migrations every time :)

punchagan:
@gs0510 The `omigrate` tool would take care of whether migrations need to run or not, running only new migrations that need to be run. Leaving this to manual discretion seems error prone to me.

gs0510:
awesome, then this looks good :) You can try it on the prod machine IST morning, shouldn't affect anyone :))

punchagan:
I've fixed the `schema_migrations` table in the production DB to allow us to run `omigrate up` in future. 

```sh
docker exec -it current-bench_pipeline_1 omigrate up --verbose --source=/app/db/migrations --database=postgresql://docker:docker@db:5432/docker
omigrate: [INFO] Version 20210013150054 has already been applied
omigrate: [INFO] Version 20210101173805 has already been applied
omigrate: [INFO] Version 20210202135643 has already been applied
omigrate: [INFO] Version 20210303145089 has already been applied
omigrate: [INFO] Version 20210920081834 has already been applied
```

This PR is now ready to be merged and deployed.

gs0510:
Awesome, thanks a ton @punchagan!

PR #195, opened 2021-11-02
Add instructions to start just frontend and DB containers

### LINKED ISSUE #107: Ability to start and run only PostgreSQL and frontend
For Multicore OCaml, we provision our hardware with specific configuration, and will re-run the experiments with different settings for compiler analysis. For such cases, we would like to use an ETL tool to just load the benchmark results and use the PostgreSQL and frontend Docker containers only for visualization.

punchagan:
Closes #107

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -112,6 +112,14 @@ pipeline container. This can be achieved by running the following command:
 make rebuild-pipeline
 ```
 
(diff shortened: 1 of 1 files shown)


gs0510:
perfect, thanks @punchagan !!

PR #197, opened 2021-11-02
Pin packages as .dev, unless they specify a version

gs0510:
When a multi-repository project adds a new package with a constraint `{= "version"}` on an existing one, this will
fail because opam will pin the new package as dev but the old package as a version of it's last release. Making the
bench-dir folder as a `dev` folder ensures that it's also picked up as a dev version.

### DIFF
--- a/pipeline/lib/custom_dockerfile.ml
+++ b/pipeline/lib/custom_dockerfile.ml
@@ -61,8 +61,13 @@ let base_dockerfile ~base =
 
 let add_workdir =
   let open Dockerfile in
-  copy ~src:[ "--chown=opam:opam ." ] ~dst:"bench-dir" ()
-  @@ workdir "bench-dir"
+  (* If the package's directory name doesn't contain a dot then opam will default to
+     using the last known version, which is usually wrong. In particular, if a multi-project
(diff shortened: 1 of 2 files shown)


pipeline/lib/custom_dockerfile.ml
```diff
@@ -61,8 +61,13 @@ let base_dockerfile ~base =
 
 let add_workdir =
   let open Dockerfile in
-  copy ~src:[ "--chown=opam:opam ." ] ~dst:"bench-dir" ()
-  @@ workdir "bench-dir"
+  (* If the package's directory name doesn't contain a dot then opam will default to
+     using the last known version, which is usually wrong. In particular, if a multi-project
+     repostory adds a new package with a constraint "{ =version }" on an existing one,
+     this will fail because opam will pin the new package as "dev" but the old one with
```
art-w, line 67: Small typos : repos**i**tory (on line 66), and the default version is `~dev` (which is even worst than `dev` because `~dev` is considered "smaller" than any other version)

art-w:
Perfect! And the comment is a nice touch! I have no idea why opam made the design decision of not having a version inside the `.opam` :/

PR #199, opened 2021-11-03
Add index make queries on (repo_id, commit) faster

punchagan:
Also, includes a commit remove an old Util to migrate some data from benchmarksrun table to benchmarks.


gs0510 (approved):
LGTM! Thanks!

PR #201, opened 2021-11-04
Pin local opam packages without confirmation for static dependencies

art-w:
The `opam pin -n .` command was introduced to fix an issue with irmin packages. I'm not entirely sure how it will interact with the computation of static dependencies, but it would help if the pins were also available at this stage. (also added the `-y` to avoid confirmation)


gs0510:
Thanks @art-w!

PR #202, opened 2021-11-04
Fix opam pin version .dev

art-w:
It seems like having the directory with a `.dev` doesn't guarantee that opam will pin all packages to the `dev` version! The setting `--with-version=dev` does the job.


gs0510 (commented):
We also need to add `@@ run "mv /usr/bin/opam-2.1 /usr/bin/opam"` because the ocaml/opam images don't come with 2.1 by default yet :)

gs0510:
Also, can we get rid of the `.dev` in the bench-dir folder that I added😂 thank you!

gs0510:
Whoops I'm sorry, I was trying to push a commit but it got closed 😨

PR #203, opened 2021-11-04
Pin to dev for opam, and undo previous dev changes

gs0510:
`ocaml/opam` doesn't come with 2.1, so need to change that. Also, just renaming the directory to `.dev` doesn't work 🤡, need to actually look at file structure, append `.dev` to file and then run opam pin. Sidestepping that by using `--with-version` (available in opam 2.1).


gs0510:
@art-w  I accidentally closed your PR, sorry about that. Merging this now which has your commits :)

PR #120, opened 2021-05-10
In progress jobs and general refactoring

rizo:
This PR focuses on improving the structure of the codebase and introduces the ability to track benchmarking job status with persistence in the database.

More precisely:

- The database schema is updated to track the job status based on the presence of job identifiers for build and run stages provided by current.
  - I might revisit this encoding in a follow up PR. Having a separate field for the status could make things easier for the frontend.
- Refactor the application entry-point (`main.ml`) to simplify config and source initialisation.
- Improve config management by introducing a centralised module for config options.
- Add migration scripts from the existing schema to the new schema with support for in-progress jobs.
  - My suggestion would be to deploy this PR as is to our testing environment, before applying the migrations.
- Introduce a shared "commit context" abstraction that precisely describes the commit the benchmarks will be run for.
- Simplify the storage module and add additional SQL statements to track job status updates.
  - I have a follow-up that reworks the db code by using Caqti.

I also suggest that we create and merge this PR into a `development` branch instead directly merging this into `main`. This would help us make and test changes in the development environment.

PR #205, opened 2021-11-05
Store metadata in a separate table

gs0510:
This PR standing on the shoulders of #120 . We have introduced a new table called `benchmarks_metadata` which allows storing only the metadata. We will eventually phase out of current table, to have one table with the metadata and another with the metrics. We also record the build_job_id and run_job_id as and when the build and run
jobs start so we can show the logs in the UI instead of showing `no data` for this pr as we do now.


pipeline/lib/pipeline.ml
```diff
@@ -231,26 +245,22 @@ let repositories = function
...
-      else pipeline ~conninfo ~run_args repository)
+      if exists ~db repository then Current.ignore_value repo
+      else pipeline ~db ~run_args repository)
     (repositories source)
 
 let v ~current_config ~docker_config ~server:mode ~(source : Source.t) conninfo
     () =
-  Db_util.check_connection ~conninfo;
-  let pipeline = process_pipeline ~docker_config ~conninfo ~source in
+  let db = new Postgresql.connection ~conninfo () in
```
art-w, line 262: I'm not sure that we want a long lived `db` connection, this can cause some issues because the database can restrict the number of active connections. In general it's better to have a short "connect/query/finish" even if it feels unoptimized! (We can look into pools if that really happens to be a bottleneck)

pipeline/lib/storage.ml
```diff
@@ -26,6 +27,7 @@ let setup_metadata ~repository (db : Postgresql.connection) =
   in
   try
     let result = db#exec query in
+    db#finish;
```
art-w, line 30: The `#finish` is a bit early before the `#get_all`! Also it needs to happen even in case of an exception :)

art-w: Maybe we should introduce a helper function:

```ocaml
let with_db ~conninfo fn =
  let db = new Postgresql.conn ... in
  match fn db with
  | v -> db#finish ; v
  | exception e ->
       Logs.err (...) ;
       db#finish ;
       raise exn
```

HACKING.md
```diff
@@ -269,6 +269,25 @@ Attaching to current-bench_db_1, current-bench_db-migrate_1, current-bench_pipel
 ...
 ```
 
+## Database schema
+
+Currently, we have two tables. We want to eventually phase out to have 
+two separate tables, one that stores the metadata and one that stores the
+metrics.
+
+The metadata table contains:
```
punchagan, line 278: I'm not sure if we really want to have a description of the DB schema in the docs. It can easily be out of sync with the actual state of the DB.

pipeline/lib/storage.ml
```diff
@@ -0,0 +1,70 @@
+let setup_metadata ~repository ~conninfo =
+  Logs.debug (fun log -> log "Inserting metada....");
```
punchagan, line 2: nitpick: Typo in metadata

pipeline/lib/storage.ml
```diff
@@ -0,0 +1,81 @@
...
+      benchmarks again for this repo and commit, so we reset the build_job_id
+      and the run_job_id.
+    *)
+    Fmt.str
+      {|
+    INSERT INTO
+    benchmark_metadata(run_at, repo_id, commit, branch, pull_number)
+    VALUES
+    (%s, %s, %s, %s, %s)
+    ON CONFLICT(repo_id, commit) DO UPDATE
```
punchagan, line 27: For this `ON CONFLICT` condition to be triggered, we'd need to add a UNIQUE constraint for `(repo_id, commit)` on the table.

pipeline/lib/pipeline.ml
```diff
@@ -151,12 +151,29 @@ let docker_make_bench ~run_args ~repository image =
         "opam exec -- make bench";
       ]
 
+let record_pipeline_stage ~stage ~serial_id ~conninfo image =
+  let+ job_id = Current_util.get_job_id image
+  and+ state = Current.state image in
+  match (job_id, state) with
```
punchagan, line 157: Isn't the intent to record metadata when the build starts or the job starts? It looks like nothing would be recorded unless there's an error in the job, given the patterns being matched here?

pipeline/lib/storage.ml
gs0510: Yes, nice catch @punchagan :) Thanks!

pipeline/lib/pipeline.ml
gs0510: So the way ocurrent works is that you can have three states an ongoing job, 
`  | Ok _ -> when the job has finished
  | Error (`Active _) -> when the job is active
  | Error (`Msg m) -> When there is an error running the job`. So we start recording when the job is active :)

punchagan: Ah, okay. Thanks for the explanation. The DB updates weren't happening because of the missing UNIQUE constraint, and I didn't correctly debug the problem. :)

punchagan (approved):
LGTM!

art-w:
Looks great to me too! ^^

PR #207, opened 2021-11-10
UI for viewing in-progress or errored runs

punchagan:
This PR builds on top of #205 to update the UI to display builds/jobs that are currently in progress or have errored out.

### DIFF
--- a/frontend/graphql_schema.json
+++ b/frontend/graphql_schema.json
@@ -1198,6 +1198,2172 @@
         "kind": "OBJECT",
         "possibleTypes": null,
         "interfaces": [],
+        "name": "benchmark_metadata",
+        "enumValues": null,
+        "description": "columns and relationships of \"benchmark_metadata\"",
+        "fields": [
+          {
+            "args": [],
+            "isDeprecated": false,
+            "deprecationReason": null,
+            "name": "branch",
+            "type": {
+              "kind": "SCALAR",
+              "name": "String",
+              "ofType": null
+            },
+            "description": null
+          },
+          {
+            "args": [],
+            "isDeprecated": false,
+            "deprecationReason": null,
+            "name": "build_job_id",
+            "type": {
+              "kind": "SCALAR",
+              "name": "String",
+              "ofType": null
+            },
+            "description": null
+          },
+          {
+            "args": [],
+            "isDeprecated": false,
+            "deprecationReason": null,
+            "name": "commit",
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              }
+            },
+            "description": null
+          },
+          {
+            "args": [],
+            "isDeprecated": false,
+            "deprecationReason": null,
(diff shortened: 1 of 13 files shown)


gs0510:
I don't know why but when I test this, the frontend fails because it gets a 404 for this request:
```
graphql-engine_1  | {"type":"http-log","timestamp":"2021-11-15T09:34:42.545+0000","level":"error","detail":{"operation":{"error":{"path":"$","error":"resource does not exist","code":"not-found"},"request_id":"a4c4ba23-f769-45cb-b2b3-577136024e82","response_size":65,"raw_query":"{\"query\":\"{\\n  allRepoIds: benchmarks(distinct_on: [repo_id]) {\\n    repo_id\\n    __typename\\n  }\\n}\\n\",\"variables\":{}}"},"http_info":{"status":404,"http_version":"HTTP/1.1","url":"/v1/grahpql","ip":"172.18.0.1","method":"POST","content_encoding":null}}}
```
This request hasn't been modified at all, so there's no reason why it should fail afaik.

punchagan:
> I don't know why but when I test this, the frontend fails because it gets a 404 for this request:

I'm not sure, either, but... 

I have been actively working on this branch and force-pushing changes. It could be that the local DB is in a weird state with the changes you made in #205 not being update in this branch, etc. 

I'm still working on some improvements to the UI and handling some edge cases. I hope to have something that's in a reviewable state by EOD or tomorrow morning.

gs0510:
Hmm, the database looks fine and I am able to make the same request from the console so I really don't know why it's failing haha. 

>I'm still working on some improvements to the UI and handling some edge cases. I hope to have something that's in a reviewable state by EOD or tomorrow morning.

Okay in that case i'll give it a rest, and see how things are tom :) Thanks!

punchagan:
image

image

image

gs0510 (commented):
Looks great! Some tiny comments, also something that @art-w and I talked about was maybe when we have the benchmarks running we can grey out the graphs, and when finished they can be "un-greyed". What do you think about this @punchagan ? :)

punchagan:
> also something that @art-w and I talked about was maybe when we have the benchmarks running we can grey out the graphs, and when finished they can be "un-greyed". What do you think about this @punchagan ? :)

Are we talking about automatically updating the UI, when the builds are finished? If so, I think we can implement that as a separate PR. Currently, I pushed a commit that greys out the graphs if they are not for the latest commit that we tried to run the benchmarks for. The user would still need to reload the page to see the status change to Passed and the "un-greyed" graphs.

punchagan:
I've updated the PR, and it's ready for another review, @gs0510!

gs0510:
Yeah I think the un-greying on reloading is fine, because I assume that's the workflow anyway. Automatically updating would mean we poll the db every x seconds to find out if things changed right?

Also, LGTM! Thanks! :))

gs0510 (approved):

punchagan:
> Automatically updating would mean we poll the db every x seconds to find out if things changed right?

Yep, something like that unless we do fancy web-sockets stuff.  I think this is good enough for now. :)

PR #211, opened 2021-11-16
Skip old commits

gs0510:
We currently re-run on all open PRs when we re-deploy CB. This means that sometimes we take more than a day to re-run all benchmarks (a lot of the older ones will fail because they don't have the changes to run the benchmarks). This patch skips all commits that are more than two weeks old.

I don't have a very strong opinion about the two weeks, we can also make it four weeks. :)

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -181,6 +181,7 @@ let github_repositories ?slack_path repo =
        Github.Api.refs api repo
   in
   let default_branch = Github.Api.default_ref refs in
+  let stale_timestamp = Util.stale_timestamp () in
   let default_branch_name = Util.get_branch_name default_branch in
   let ref_map = Github.Api.all_refs refs in
   let+ _, repo = repo in
@@ -189,12 +190,15 @@ let github_repositories ?slack_path repo =
     (fun key head lst ->
       let commit = Github.Api.Commit.id head in
       let repository = repository ~commit ~github_head:head in
-      match key with
-      (* Skip all branches other than master, and check PRs *)
-      | `Ref branch when branch = default_branch ->
-          repository ~branch:default_branch_name () :: lst
-      | `Ref _ -> lst
-      | `PR pull_number -> repository ~pull_number () :: lst)
+      (* If commit is more than two weeks old, then skip it.*)
+      if Github.Api.Commit.committed_date head > stale_timestamp then
+        match key with
+        (* Skip all branches other than the default branch, and check PRs *)
+        | `Ref branch when branch = default_branch ->
+            repository ~branch:default_branch_name () :: lst
+        | `Ref _ -> lst
+        | `PR pull_number -> repository ~pull_number () :: lst
+      else lst)
     ref_map []
 
 let repositories = function
--- a/pipeline/lib/util.ml
+++ b/pipeline/lib/util.ml
@@ -6,3 +6,14 @@ let get_branch_name branch =
   let prefix = "refs/heads/" in
   let len_prefix = String.length prefix in
   String.sub branch len_prefix (String.length branch - len_prefix)
+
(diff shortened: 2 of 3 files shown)


pipeline/lib/util.ml
```diff
@@ -6,3 +6,14 @@ let get_branch_name branch =
   let prefix = "refs/heads/" in
   let len_prefix = String.length prefix in
   String.sub branch len_prefix (String.length branch - len_prefix)
+
+let current_timestamp () =
```
punchagan, line 10: `current_timestamp` seems a bit misleading. Should we rename this to something like `stale_commit_timestamp`?

pipeline/lib/util.ml
```diff
@@ -6,3 +6,14 @@ let get_branch_name branch =
...
+  let two_weeks_in_unix_epoch = 1209600. in
+  let curr_gmt = Unix.gmtime (Unix.time () -. two_weeks_in_unix_epoch) in
+  let year = string_of_int curr_gmt.tm_year in
+  let month = string_of_int curr_gmt.tm_mon in
+  let day = string_of_int curr_gmt.tm_mday in
+  let hour = string_of_int curr_gmt.tm_hour in
+  let min = string_of_int curr_gmt.tm_min in
+  let sec = string_of_int curr_gmt.tm_sec in
+  year ^ "-" ^ month ^ "-" ^ day ^ "T" ^ hour ^ ":" ^ min ^ ":" ^ sec ^ "Z"
```
punchagan, line 19: I wish there was a better way to format timestamps in OCaml. :cry: 

There's a bunch of things that need to be taken care of here, looking at the docs for [`Unix.tm`](https://ocaml.org/api/Unix.html#TYPEtm), the code needs `year` to add `1900` and `month` to add `1`. 

Also, we need all the numbers to be zero padded. 

```ocaml
  let year = curr_gmt.tm_year + 1900 in
  let month = curr_gmt.tm_mon + 1 in
  let day = curr_gmt.tm_mday in
  let hour = curr_gmt.tm_hour in
  let min = curr_gmt.tm_min in
  let sec = curr_gmt.tm_sec in
  Format.sprintf "%04d-%02d-%02dT%02d:%02d:%02dZ" year month day hour min sec
```

Would something like this be better?

pipeline/lib/util.ml
```diff
@@ -6,3 +6,14 @@ let get_branch_name branch =
   let prefix = "refs/heads/" in
   let len_prefix = String.length prefix in
   String.sub branch len_prefix (String.length branch - len_prefix)
+
+let current_timestamp () =
+  let two_weeks_in_unix_epoch = 1209600. in
```
punchagan, line 11: This is really `two_weeks_in_seconds`. :)

pipeline/lib/pipeline.ml
```diff
@@ -181,6 +181,7 @@ let github_repositories ?slack_path repo =
        Github.Api.refs api repo
   in
   let default_branch = Github.Api.default_ref refs in
+  let current_timestamp = Util.current_timestamp () in
```
punchagan, line 184: The name `current_timestamp` is misleading here. May be something like `stale_commit_timestamp` would work better?

punchagan (changes requested):
Thanks for working on this, @gs0510. The idea looks good to me, in general. I have a few comments on the implementation.

punchagan:
> I don't have a very strong opinion about the two weeks, we can also make it four weeks. :)

Is it worth making this a config variable or is it not worth the extra hassle?

gs0510:
> Is it worth making this a config variable or is it not worth the extra hassle?

I don't know, I think it should be okay like this 🤷🏾 If someone complains we can think about it 😁 

Also thanks for the review! :))

punchagan (approved):

PR #213, opened 2021-11-23
Add OCluster to development docker

art-w:
This is preliminary integration with ocluster: This PR only adds an ocluster to the development docker, but it doesn't actually change the pipeline to use workers. It can be helpful to play with workers though, and more integration will follow soon!

Some things to note:
- The `capnp-secrets` contains the permissions to talk with the ocluster. The `admin.cap` is used by the pipeline in `entrypoint.sh` to generate `submission.cap` which can then be used to submit new jobs to the cluster.
- The workers need access to `pool-linux.cap` to receive jobs from the cluster (for that pool, you can add more, the `linux` name is arbitrary). By default, this file has `cluster:9000` as the host... so you may need to edit it to a real IP or `localhost:9000` or add `cluster` to your hosts.
- The `local-test-repo` should be specified in the `development.env` as an absolute path: Otherwise when the pipeline will submit jobs to the cluster, the workers will not be able to find the git repository (since they are running outside of the pipeline docker, the external path should be the same and not `/app/local-test-repo`).

See the [OCluster readme](https://github.com/ocurrent/ocluster#workers) for some commands to experiment with workers!


gs0510:
LGTM! I have one question! Will it be possible to add two modes of starting the local dev mode? One cluster and one without one? What do you think about this idea? :)

art-w:
I am a bit conflicted: It would be convenient to ease development, but would also require to maintain two different build-and-run setup (one in ocurrent for the pipeline, and one for the worker).

art-w:
Redundant and less uptodate than https://github.com/ocurrent/current-bench/pull/216

PR #214, opened 2021-11-24
Hot reload pipeline on code changes for quick development

art-w:
When developing, the pipeline restart can be very slow! The `reload.sh` scripts watches for changes and re-run the pipeline on code changes (without rebuilding a fresh new docker every time).

I'm not sure if we can get rid of `Dockerfile.dev`: it's very similar to the production one, but must keep the opam environment. Do you know a better way?

### DIFF
--- a/environments/development.docker-compose.yaml
+++ b/environments/development.docker-compose.yaml
@@ -61,6 +61,7 @@ services:
   pipeline:
     build:
       context: ../pipeline
+      dockerfile: ./Dockerfile.dev
       args:
         TARGET_ARCH: ${OCAML_BENCH_TARGET_ARCH?required}
     volumes:
@@ -69,9 +70,11 @@ services:
       - ../local-test-repo:/app/local-test-repo
       - /var/run/docker.sock:/var/run/docker.sock
       - ../pipeline/db/migrations:/app/db/migrations
+      - ../pipeline:/mnt/project
     ports: ["${OCAML_BENCH_PIPELINE_PORT?required}:${OCAML_BENCH_PIPELINE_PORT?required}"]
     command:
-      - "current-bench-pipeline"
+      - "/mnt/project/reload.sh"
+      - "/app/bin/current-bench-pipeline"
       - "local"
       - "/app/local-test-repo"
       - "--verbose"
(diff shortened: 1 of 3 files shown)


gs0510:
@art-w What do you mean when you say the dockerfile must keep the opam environment?

art-w:
The `pipeline/Dockerfile` used in production has this line:

```
FROM debian:stable-slim
```

But this removes the opam libraries required to re-build the sources when there is a change :/

gs0510 (approved):
Ah I see! Yes it's a bit unfortunate that we have to duplicate it, but should be okay :)

punchagan:
I just noticed that `dune build` has a `--watch` mode. @art-w I was curious to know why you chose not to use that and go with a custom `inotifywait` call, instead. Could you explain why you took this approach? :)

art-w:
I forgot about `dune build --watch`! We would still need to `inotifywait` on the binary to know when to restart it though?.. (but at least we wouldn't need to list all the source files)

punchagan:
Yeah, it does sound like we would need the `inotifywait` to restart the binary, and leaving the list of source files to watch to dune seems like a nice idea. Should I try to get this working and submit a PR, @art-w ?

art-w:
Yes sure! It might be challenging to output the build logs, and I don't know which libraries `dune` requires to enable inotify (otherwise I think it fallbacks to slowly retrying the build every second)

gs0510:
dune also uses `inotifywait` to implement the watch mode, so as long as you have that I think it should be okay!

punchagan:
> dune also uses `inotifywait` to implement the watch mode, so as long as you have that I think it should be okay!

Yes, that seems to be correct. `dune` seems to be running some thing like this, which is pretty similar to what we do in our `inotifywait` call except, it also listen for `delete` events. 

```sh
/usr/bin/inotifywait -r . --exclude /_build|/_opam|/_esy|/\..+|~$|/#[^#]*#$|4913 -e close_write -e delete --format %w%f -m -q
```

I pushed an experimental commit [here](https://github.com/punchagan/current-bench/commit/382dd68b2367e0d53789c8cf60547ac3e273625b) that seems to do what we need. It doesn't look better than what we already have, but leaves the burden of the correct `inotifywait` command to the dune devs.  Is it worth sending a PR with this commit?

punchagan:
I've opened #227  for whatever it's worth. Feel free to close it, if it doesn't feel like a real improvement. :)

PR #215, opened 2021-11-24
Extract json from mixed output

art-w:
This PR is a rebase of the work of @shubhamkumar13 which allow to extract the json metrics when the `make bench` output also contains other things. This will be especially important with ocluster workers, where stderr and stdout will be interleaved (at least for the first version).

### DIFF
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -29,6 +29,7 @@ depends: [
   "postgresql"
   "rresult"
   "omigrate"
+  "pcre"
 ]
 
 pin-depends: [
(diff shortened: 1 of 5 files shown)


gs0510 (changes requested):
I think you need to add `pcre` to `current-bench.opam` for the CI to pass. Otherwise looks great! Thanks @art-w!

art-w:
Good catch! All the credits go to @shubhamkumar13 :)

gs0510 (approved):

PR #216, opened 2021-11-25
 Add a custom OCluster worker build-and-run-benchmarks 

art-w:
This PR adds an ocluster worker that can be used to build and run benchmarks from the main pipeline. It also adds a default worker to the development docker compose so that everything continue to work out of the box.

There are still a lot of things to do before this PR can be merged:
- Fix the production, as workers/ocluster are not enabled there! (and it will break production if released)
- Fix the "multicore repositories" settings, currently the worker doesn't have access to the repository name for no good reason.
- The build and run are not cleanly separated, resulting in a single job_id and job log file.
- Clean up a few things, like the duplication of the `aslr_seccomp.json` file and `Docker_config`
- Remove the dependency on my branch of `ocluster`, I will push a PR to them... but I am having some issues with their obuilder git submodule and the current solution is not great.


art-w:
- Production should work now!
- We will use the workers pool as a way to dispatch repositories to the same configuration every time. The workers should not apply special settings based on the repository name.
- The build and run logs are still not cleanly separated, but there is no easy way to do it at the moment.
- `Docker_config` was removed from the pipeline. We should remove the use of docker entirely in the pipeline and do the opam dependency analysis like ocaml-ci (which will allow us to remove the duplicate `aslr_seccomp.json`)

I think the remaining TODOs are independent enough to be handled as separate PRs. If you have some time, can you confirm that the cluster and workers are working correctly?

gs0510 (approved):
Thanks for this @art-w! 🎉  This looks and works great! :)) Sorry for the delay in my review!

punchagan:
> The build and run logs are still not cleanly separated, but there is no easy way to do it at the moment.

Do you have any ideas on how to do this? Currently the UI shows two separate links for build logs and execution logs, which both point to the same log file. It may be fine for now, but we could tweak the UI to show only one link if both the log files are the same.

gs0510:
@punchagan Yes that's a reasonable solution (showing only one link on the UI); the current workflow would be that we put the same link on both the build and the run jobs, but your tweak sounds better :)

PR #222, opened 2021-11-29
Fix broken frontend watcher screen command

punchagan:
The screen command that runs the `yarn watch` command incorrectly used the `-L`
argument for the log file, instead of the `-Logfile` argument. Not sure how
this has been working so far. Possibly, `node:lts` image changed the base
image, or something which changed the screen version, but I couldn't confirm
this hypothesis.


gs0510 (approved):
ah nice catch, thanks @punchagan!

PR #223, opened 2021-11-29
Add -L to screen

gs0510:
The fix in #222 was incomplete because we need to turn on the logging mode and then set the logfile path for screen. :)


punchagan:
Oh, wow! Thanks!

Issue #126, opened 2021-05-18
PR titles are missing


emillon:
Hi,

The UI renders just the PR numbers, which can be difficult to identify:

![image](https://user-images.githubusercontent.com/496345/118638460-27e36e00-b7d7-11eb-9f3c-90065e4f4661.png)

I would suggest adding the PR titles like this:

![image](https://user-images.githubusercontent.com/496345/118639085-cc65b000-b7d7-11eb-86ef-f33539502799.png)

Maybe just in the top part because it will be difficult to render the full title in the left bar.

gs0510:
This is a good idea, thanks @emillon!

gs0510:
Adding some notes from weekly meeting: `ocurrent` doesn't support this, we'd likely want to use the ocurrent Github API for this since that can get the webhook request and the title from the request. Something to keep in mind would also be the fact that PR title can change, so the data in the database would need to be updated and the frontend would need to poll those changes.

PR #224, opened 2021-11-29
Add PR titles to UI

gs0510:
Ocurrent fetches all the refs and PRs using the github graphql interface. The interface is abstracted out so we can get the title
making another graphql query to github. It is not possible to just make one query because all the types are not exposed by OCurrent, so we would need to do a lot of copying of types if we wanted to avoid making another query. The frontend then displays the title if it's available in the benchmark_metadata table.

### DIFF
--- a/frontend/graphql_schema.json
+++ b/frontend/graphql_schema.json
@@ -1,314 +1,247 @@
 {
   "__schema": {
-    "directives": [
-      {
-        "args": [
-          {
-            "name": "if",
-            "defaultValue": null,
-            "type": {
-              "kind": "NON_NULL",
-              "name": null,
-              "ofType": {
-                "kind": "SCALAR",
-                "name": "Boolean",
-                "ofType": null
-              }
-            },
-            "description": null
-          }
-        ],
-        "name": "include",
-        "locations": [
-          "FIELD",
-          "FRAGMENT_SPREAD",
-          "INLINE_FRAGMENT"
-        ],
-        "description": null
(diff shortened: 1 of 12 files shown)


gs0510:
Screenshot from 2021-11-29 17-37-20
)

gs0510:
![Screenshot from 2021-11-29 17-37-20](https://user-images.githubusercontent.com/8718814/143907094-1064967b-71ee-4cb0-a208-90393d6b6e91.png

gs0510:
Screenshot from 2021-11-29 17-37-20

art-w:
Very nice! As a next step, this PR will allow us to update the sidebar as the preview proposed in https://github.com/ocurrent/current-bench/issues/126 was looking really good!

I think the current code doesn't update the missing PR titles for already processed commits (because the `pr_title` only gets inserted during `setup_metadata` if the `exists` test failed).

gs0510:
Ah nice! I will edit the frontend to look like what's proposed in #126 (it looks nicer imo), maybe we don't care about the sidebar, it doesn't look like it'd be very useful, what are your thoughts @art-w ? :)

art-w:
I think it would be nice to have the PR titles in the sidebar, as they would help navigation :)

gs0510:
Screenshot from 2021-11-30 18-11-51
You can now see the pr title on the sidebar!

art-w:
LGTM! Thanks!

PR #227, opened 2021-12-01
Use dune build -watch instead of custom inotifywait command

punchagan:
Let dune build take care of the source files to watch, instead of listing all
the files manually in our reload.sh script

### DIFF
--- a/pipeline/reload.sh
+++ b/pipeline/reload.sh
@@ -2,27 +2,30 @@
 
 COMMAND="${*@Q}"
 
+DIR=_build/default/bin
+EXE="${DIR}/main.exe"
+
 function run {
   cd /app
   sh -c "$COMMAND" &
+  cd /mnt/project
 }
 
 run
 
-cd /mnt/project
 /mnt/project/dev/github-app.sh
 
 while :
 do
-  cd /mnt/project
-
-  inotifywait -q -q -e CLOSE_WRITE \
-    lib/dune lib/*.ml lib/*.mli \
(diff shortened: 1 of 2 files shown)


gs0510 (approved):
thanks @punchagan, lgtm but will wait for @art-w to review :)

pipeline/reload.sh
```diff
@@ -2,24 +2,29 @@
...
-    bin/dune bin/*.ml bin/*.mli
-
-  dune build bin/main.exe \
+  # NOTE:
+  # 1. inotifywait exits with a failure status code when main.exe gets deleted
+  #    before a rebuild. We want to exit early, when that happens.
+  # 2. Also, we sleep for a short while, to ensure the process has been killed
+  #    before copying the new executable and trying to run it.
+  test -f "${EXE}" \
+  && inotifywait -q -q -e CLOSE_WRITE "${EXE}" \
```
art-w, line 25: I'm a tiny bit worried about what happens when the executable is removed / not there. For example if the first `dune build` fails, then the while loop will test a bit too fast for the executable existence. Can we perhaps use inotifywait to wait for the executable to exists? or add a sleep somewhere?

It looks great otherwise, thank you!

punchagan:
I'm starting to feel like what we have on `main` is more "stable" since we run the dune build "synchronously" while trying to restart the pipeline. I pushed a commit that should work slightly better than my previous commit, though.

art-w:
It works great! Thanks :)

gs0510:
thanks @punchagan!

PR #228, opened 2021-12-01
Fix pipeline deployability testing

art-w:
The deployability testing seems to fail because it executes the `pipeline/Dockerfile` at the root of the project, and not in the subfolder `pipeline`. Let's see if it works now...

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,4 @@
+.git*
 frontend/node_modules
+current-bench-bechamel
+scripts
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -4,6 +4,8 @@ maintainer: ["Rizo I. <rizo@tarides.com>" "Gargi Sharma <gargi@tarides.com>"]
 authors: ["Rizo I. <rizo@tarides.com>" "Gargi Sharma <gargi@tarides.com>"]
 homepage: "https://github.com/ocurrent/current-bench"
 bug-reports: "https://github.com/ocurrent/current-bench/issues"
+dev-repo: "git+https://github.com/ocurrent/current-bench.git"
+
 depends: [
   "ocaml"   {>= "4.13.0"}
   "dune" {>= "2.0"}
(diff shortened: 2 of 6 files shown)


gs0510:
@art-w this is something i have looked into and have an idea about what's wong; will talk more about it tomorrow :)

.dockerignore
```diff
@@ -1 +1,3 @@
 frontend/node_modules
+current-bench-bechamel
+scripts
```
gs0510, line 3: why did you ignoring the scripts ? :)

gs0510 (approved):
one question, otherwise LGTM! Thanks!

art-w:
It works!.. I'm not sure what we can do about capnp on Alpine 3.13. Can we tell ocaml-ci that we are not interested by that platform?

gs0510:
@art-w can you build the images locally with the make targets? I remember i broke those when i was trying to fix the root problem...

.dockerignore
art-w: Because I wanted to relaunch the deployability testing and needed to update the commit to do so! :D The failure was not our fault, but a network issue on their end.
I believe this `.dockerignore` is only used by the CI to copy the whole repository to docker -- and we only use the `scripts/` directory for human convenience, so docker doesn't need it to build our images. It doesn't really matter given the size of that directory though!

gs0510: haha okay, that sounds reasonable :)

art-w: It was a good question! I added the `.git` directory to the `.dockerignore` since it's a lot more expensive to copy :)

gs0510:
@art-w from the oclc-ci channel, we can add this to the opam file to avoid alpine...

```
available: [ <filter> ]: can be used to add constraints on the OS and other global variables. In case the filter doesn't evaluate to true, the package is disabled.

This field is evaluated before request solving or any actions take place ; it can only refer to global variables, since it shouldn't depend on the current switch state. An unavailable package won't generally be seen on the system, except with opam list -A.
```

art-w:
Yes, it still works locally! I'm going to experiment with the opam `available` field to ignore alpine, I'm not sure what format is expected.

art-w:
:heavy_check_mark: :heavy_check_mark: :heavy_check_mark: !

gs0510:
Woohooo! Thanks @art-w!!

PR #229, opened 2021-12-02
Support for opam pin-depends and depexts

art-w:
The current strategy for optimizing opam dependencies during the docker build has some limitations: no support for system dependencies and opam pin-depends. It also requires to run docker just for the analysis.

This PR uses a simpler tactic: we copy only the `.opam` files in the docker, pin them and run the installations. If the opam files did not change, then the docker cache will be able to reuse the previous result.

The code had some special support for `*-bench.opam` files, but looking at irmin for example, their `-bench.opam` requires the other packages to be present (so we have to copy and pin everything anyway).

### DIFF
--- a/pipeline/lib/custom_dockerfile.ml
+++ b/pipeline/lib/custom_dockerfile.ml
@@ -50,79 +50,31 @@ let get_all_files ~repository =
   |> let> commit = Repository.src repository in
      Cache.get () { Get_files.Key.commit }
 
-let base_dockerfile ~base =
-  let open Dockerfile in
-  from (Docker.Image.hash base)
-  @@ run
-       "sudo apt-get update && sudo apt-get install -qq -yy libffi-dev \
-        liblmdb-dev m4 pkg-config gnuplot-x11 libgmp-dev libssl-dev \
-        libpcre3-dev && opam remote add origin https://opam.ocaml.org && opam \
-        update"
-  @@ run "sudo mv /usr/bin/opam-2.1 /usr/bin/opam"
-
-let add_workdir =
-  let open Dockerfile in
-  copy ~src:[ "--chown=opam:opam ." ] ~dst:"bench-dir" ()
-  @@ workdir "bench-dir"
-  @@ add ~src:[ "--chown=opam ." ] ~dst:"." ()
-  @@ run "opam exec -- opam pin -y -n --with-version=dev ."
-
-let minimal_dockerfile ~base =
-  let open Dockerfile in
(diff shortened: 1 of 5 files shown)


pipeline/lib/custom_dockerfile.ml
```diff
@@ -50,92 +50,40 @@ let get_all_files ~repository =
...
-  @@ install_static_dependencies
-  @@ add_workdir
-  @@ run "%s" (opam_install ~opam_file)
-
-let dockerfile ~pool ~run_args ~repository ~opam_file =
-  let* dependencies =
-    discover_dependencies ~pool ~run_args ~repository ~opam_file
-  in
-  with_base (dockerfile ~dependencies ~opam_file)
+let dockerfile ~base = Dockerfile.crunch (dockerfile ~base)
```
gs0510, line 75: What does `crunch` do? :)

pipeline/lib/custom_dockerfile.ml
```diff
@@ -50,92 +50,40 @@ let get_all_files ~repository =
...
+  @@ run "opam remote add origin https://opam.ocaml.org"
+  @@ run "opam update"
   @@ workdir "bench-dir"
-  @@ add ~src:[ "--chown=opam ." ] ~dst:"." ()
+  @@ copy ~src:[ "--chown=opam:opam ./*.opam" ] ~dst:"." ()
   @@ run "opam exec -- opam pin -y -n --with-version=dev ."
+  @@ run "opam exec -- opam install -y --depext-only ."
+  @@ run "opam exec -- opam install -y --deps-only ."
+  @@ copy ~src:[ "--chown=opam:opam ." ] ~dst:"bench-dir" ()
+  @@ add ~src:[ "--chown=opam ." ] ~dst:"." ()
```
gs0510, line 73: Should this also be `--chown=opam:opam`? :)

art-w: Yes indeed!

pipeline/lib/custom_dockerfile.ml
art-w: `Dockerfile.crunch` collapses multiple docker `run` commands into a single one, using `&&` in between (so that the docker cache a bit smaller). It's a prettier alternative to writing long commands in multiline strings :P

gs0510: nice!! very cool!

gs0510 (approved):
LGTM! Thanks @art-w!

PR #231, opened 2021-12-03
Correctly handle displaying information about cancelled builds in the UI

punchagan:
The PR adds new columns to indicate cancelled builds (and the reason) in benchmarks_metadata table. It also updates the UI to display cancelled build status correctly.  

The PR doesn't yet update the GitHub status about cancelled builds. When a PR is merged while the benchmarks are running, the status of the check would end-up staying as running in the GitHub UI, instead of saying something like cancelled. This may be alright for now, but would be nice to fix. I'll try to work on this as a follow-up to this PR, if it makes sense to do so. 

Thanks to @gs0510 and @art-w for helping me with ideas on how to go about working on this PR. 

image

### DIFF
--- a/frontend/graphql_schema.json
+++ b/frontend/graphql_schema.json
@@ -1,247 +1,256 @@
 {
   "__schema": {
+    "directives": [
+      {
+        "args": [
+          {
+            "name": "if",
+            "defaultValue": null,
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Boolean",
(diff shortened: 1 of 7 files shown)


punchagan:
(Looks like the benchmarks ci fails because the dev server doesn't have enough space: http://147.75.32.19:8081/job/2021-12-03/113337-docker-build-fa2a00 )

gs0510:
LGTM @punchagan! Great work! :)) I'll give it one more pass in the evening before merging :)

gs0510:
Thanks @punchagan 🎉

PR #233, opened 2021-12-06
Fix default dockerfile when .opam is missing

art-w:
Docker fails on `COPY ./*.opam .` when there are no `.opam` files.


gs0510 (approved):

CraigFe:
Looks like this change is causing trouble for Irmin's benchmarks (e.g. [here](http://autumn.ocamllabs.io:8081/job/2021-12-06/164726-docker-build-cec218)):

```
Step 4/7 : COPY --chown=opam:opam ./*.opam .
When using COPY with more than one source file, the destination must be a directory and end with a /
```

Perhaps the solution is just to copy into `./` instead?

gs0510:
nice catch, thanks for the report @CraigFe! 💜

PR #238, opened 2021-12-07
Unset failed flag when recording stage start

### LINKED ISSUE #237: Status not handled correctly
For http://autumn.ocamllabs.io/mirage/irmin/pull/1653, we run the benchmarks succesfully and display them but we still get a "failed" on the status on the UI.
Screenshot from 2021-12-06 19-28-45

FWIW the benchmarks did fail before, so I think we just forget to set the failed status to false when they eventually run successfully.

punchagan:
Multiple builds could run for the same commit -- we allow manual rebuilds and
also run builds for latest failed commits, when restarting current-bench. So, a
previously failed build could succeed in future. This commit explicitly unsets
failed to false when starting a specific stage.

Closes #237

### DIFF
--- a/pipeline/lib/storage.ml
+++ b/pipeline/lib/storage.ml
@@ -26,7 +26,7 @@ let setup_metadata ~repository (db : Postgresql.connection) =
     VALUES
     (%s, %s, %s, %s, %s, %s)
     ON CONFLICT(repo_id, commit) DO UPDATE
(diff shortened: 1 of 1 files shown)


art-w:
Could we set the `failed=false` in `setup_metadata`, just like we erase the build and job id there?

gs0510 (approved):
thanks @punchagan!

PR #243, opened 2021-12-10
Don't show build logs link if it is same as job execution logs

punchagan:
https://github.com/ocurrent/current-bench/pull/216 added an OCluster worker to
build and run benchmarks, and the logs for build and execution are in a single
file. This commit shows a single link in the UI when there's a single log file.


gs0510 (approved):
LGTM! Thanks @punchagan!

PR #250, opened 2021-12-14
Extract json from worker logs

art-w:
The current way of extracting the json metrics requires to read the whole build-and-run logs from the worker in memory, before we can proccess it. This PR does the log read chunk by chunk, and keeps only the json-looking parts in memory (with a custom pseudo parser to identify the json parts).

- Fix the redundant writes to the database by using `Current_cache.Output`
- Remove the dependency to `Pcre`... it segfaulted in production!

In the future, it would be nice to write the metrics to the database as soon as we read them from the log (rather than wait until `make bench` terminates). This would allow for early feedback on the frontend on how the benchmark is going.

### DIFF
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -36,7 +36,6 @@ depends: [
   "postgresql"
   "rresult"
   "omigrate"
-  "pcre"
 ]
 
 pin-depends: [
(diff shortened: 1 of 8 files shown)


gs0510 (commented):
this looks great!! I'll give it one last pass tomorrow, and also let other take a look before merging :))

gs0510 (approved):

PR #252, opened 2021-12-15
Make the debian version more explicit.

tmcgilchrist:
Suggestion to make this version more explicit.


gs0510:
thanks @tmcgilchrist!

PR #253, opened 2021-12-15
Fix bug in computeStdDev when only a single commit has been recorded

punchagan:
We perform a division by 0, when we try to calculate the standard deviation for
an array with a single value in it.


gs0510 (approved):
😬😬 nice catch, thanks @punchagan!!

PR #254, opened 2021-12-15
Allow setting a description for the metrics

punchagan:
image

### DIFF
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -66,7 +66,12 @@ module V2 = struct
 
   type value = Float of float | Floats of float list
 
-  type metric = { name : string; value : value; units : string }
+  type metric = {
+    name : string;
+    description : string;
+    value : value;
+    units : string;
+  }
 
   type result = { test_name : string; metrics : metric list }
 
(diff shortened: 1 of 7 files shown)


pipeline/lib/current_bench_json.ml
```diff
@@ -177,8 +187,8 @@ let of_json json =
   | _ -> V1.to_v2 @@ V1.of_json json
 
 let validate t =
-  let open V2 in
   let tbl = Hashtbl.create 16 in
+  let open Latest in
```
gs0510, line 191: Where does the `Latest` come from? :)

gs0510 (commented):
LGTM, just one question! thanks @punchagan!

pipeline/lib/current_bench_json.ml
punchagan: We assign `module Latest = V2` just a few lines above, in this file. We could define a `V3` in future, and `validate` shouldn't need to be changed at that point, with this change.

gs0510 (approved):
thanks @punchagan!

pipeline/lib/current_bench_json.ml
gs0510: great, thanks! :))

PR #255, opened 2021-12-16
Display current bench version info in the UI

punchagan:
It is some times useful to see the version of the current bench that is running in the UI. This PR displays the commit hash of the latest commit of the current bench repository in the UI.


gs0510 (approved):
this is a good idea, thanks @punchagan!

PR #256, opened 2021-12-16
Add validations for docker compose env files

punchagan:


scripts/validate-env.sh
```diff
@@ -0,0 +1,86 @@
...
+    then
+        echo "$1 is valid"
+    else
+        echo "$1 is NOT valid"
+        ERRORS="yes"
+    fi
+}
+
+is_graphql_url() {
+    regex='https?://.*/v1/graphql'
```
gs0510, line 42: This also 😿

scripts/validate-env.sh
```diff
@@ -0,0 +1,86 @@
...
+    if [ -d $1 ]; then
+        echo "$1 exists"
+    else
+        echo "${1} is NOT a valid directory"
+        ERRORS="yes"
+    fi
+}
+
+url_with_schema() {
+    regex='https?://.*'
```
gs0510, line 31: this should be `http` 😬

gs0510 (changes requested):
looks great thanks @punchagan!!

scripts/validate-env.sh
punchagan: `s?` would mean that `s` either appears, or not.

gs0510 (approved):

PR #257, opened 2021-12-16
Config repositiories to run with specific workers and ocaml versions

art-w:
This PR adds a static configuration for the repositories, which allows to specify on which worker(s) and with which ocaml version(s) the benchmark should run. The main goal is to produce deterministic runs as we don't want the metrics to come from different hardware setup, as it would render the graphs unusable! It also makes it possible to specify which repositories should run on the "sandmark" workers (with multicore enabled etc).

In `environments/development.conf` and `production.conf`, you can specify a json list of repo configurations:

```json
[
  {
    "name": "author/repo",
    "worker": "autumn",
    "image": "ocaml/opam"
  },
  {
    "name": "local/local",
    "image": "ocaml/opam:debian-ocaml-4.11"
  }
]
```

If unspecified, the default behavior is to run on the `autumn` worker with the base image `ocaml/opam` (if you omit worker or image from the conf then it also takes the default value). If a `bench.Dockerfile` is present in the repo, then the configuration base image is ignored.

The same repository can be repeated with different configurations: This allow to run the same benchmark with different ocaml versions (or different hardware if you vary the worker.) On the frontend there's a new select in the sidebar to inspect the different runs... this could be improved with graph overlays in the future :)

I had to remove the slack support because the current behavior was to send the whole output log as a slack message... This is potentially huge (and even worse if you run it multiple times with different ocaml versions), and as a result it is not used in production. We will have to create a "summary" message if we really want to support slack!


pipeline/lib/pipeline.ml
```diff
@@ -208,19 +210,19 @@ let exists ~conninfo repository =
   let db = new Postgresql.connection ~conninfo () in
   let exists = Benchmark.Db.exists db repository in
   db#finish;
-  exists
+  false && exists
```
gs0510, line 213: Why do we need an && with `false` here?

gs0510 (commented):
Some tiny nits, otherwise LGTM! Thanks @art-w :))

gs0510:
I also have a question re. opam dependencies, do you think we should display in the side drop down only the benchmarks that pass? Should we say the benchmarks failed if they fail on some versions or should we have another status (something like `partial`)?

Also, if it's okay, I'd like to work on the opam dependency magic :D

pipeline/lib/pipeline.ml
art-w: Oups no we don't!

pipeline/Dockerfile
```diff
@@ -59,3 +59,4 @@ COPY --from=build /home/opam/.opam/4.13/bin/omigrate /app/bin/omigrate
 COPY --from=build /home/opam/.opam/4.13/bin/ocluster-admin /app/bin/ocluster-admin
 COPY --from=build /mnt/project/_build/default/bin/main.exe /app/bin/current-bench-pipeline
 COPY pipeline/entrypoint.sh /app/entrypoint.sh
+COPY environments/production.conf /app/production.conf
```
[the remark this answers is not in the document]
art-w: The `development.docker-compose.yaml` already has access to the `development.conf` in `/mnt/environments` so no need to copy it again :)

environments/development.conf
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "name": "local/local"
+  },
+  {
+    "name": "local/local",
+    "image": "ocaml/opam:debian-ocaml-4.11"
```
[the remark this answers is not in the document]
art-w: No reason at all, I changed it to 4.13 :)

environments/production.conf
```diff
@@ -0,0 +1 @@
+[]
```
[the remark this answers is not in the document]
art-w: It depends on whether we want the production configuration to be public or not... currently it's not, so I could change this to a `.template` and add a configuration variable to point to a secret `production.conf`?

art-w:
> Also, if it's okay, I'd like to work on the opam dependency magic :D

Yes of course, please do!

The UI in this PR is not very great with respect to displaying the different workers/ocaml versions (a simple select dropdown in the sidebar... and the docker image's names are too long). Ideally it would be great to have a list of runs (with perhaps checkboxes to select which runs to show simultaneously as graph overlays)

- On current-bench website, I think all the runs should be visible since that's how you get to the build/run logs to debug the issue! But we could add icons :x: near the run's name to hint at a failure (currently you have to cycle manually between the runs to discover the failing ones) 
- For the github feedback, apparently there are lots of [possible outcomes](https://docs.github.com/en/rest/reference/repos#statuses?query=statuses) in the `conclusion`: action_required, cancelled, failure, neutral, success, skipped, stale, or timed_out (but no `partial` failure?)

pipeline/lib/pipeline.ml
```diff
@@ -210,17 +212,17 @@ let exists ~conninfo repository =
...
     (fun repo ->
       let* repository = repo in
       if exists ~conninfo repository
       then Current.ignore_value repo
-      else pipeline ~ocluster ~conninfo repository)
+      else pipeline ~config ~ocluster ~conninfo repository)
     (repositories sources)
 
-let v ~current_config ~server:mode ~sources conninfo () =
+let v ~config ~current_config ~server:mode ~sources conninfo () =
```
gs0510, line 225: Can we rename `config`? Having both `config` and `current_config` as parameters looks a bit confusing :)

gs0510 (commented):
One last nit @art-w, otherwise this is good to go! thank you!!

pipeline/lib/pipeline.ml
art-w: Indeed! `current_config` is not actually used so it was removed :)

gs0510 (approved):

PR #258, opened 2021-12-17
Capture json validation errors in the DB

punchagan:
JSON validation silently fails, and this PR captures the errors to the DB and displays them in the UI

image

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -76,7 +76,7 @@ let record_pipeline_stage ~stage ~serial_id ~conninfo image job_id =
       Storage.record_stage_start ~stage ~job_id ~serial_id ~conninfo
   | Some _, Error (`Msg m) ->
       Logs.err (fun log -> log "Error in %s stage: \n%s\n" stage m);
-      Storage.record_stage_failure ~stage ~serial_id ~conninfo
+      Storage.record_stage_failure ~stage ~serial_id ~reason:m ~conninfo
   | _ -> ()
 
 module Env = Custom_dockerfile.Env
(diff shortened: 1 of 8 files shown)


pipeline/lib/pipeline.ml
```diff
@@ -121,11 +121,20 @@ let pipeline ~ocluster ~conninfo ~repository =
   and+ () =
     record_pipeline_stage ~stage:"run_job_id" ~serial_id ~conninfo worker
       worker_job_id
-  and+ output = Json_stream.save ~conninfo ~repository worker_job_id in
+  in
+  let output = Json_stream.save ~conninfo ~repository worker_job_id in
```
art-w, line 125: I think you can define this `output` before the `let+`, and have a big chain of `and+` :

```ocaml
let output = Json_stream.save ~conninfo ~repository worker_job_id in
let+ () = record_pipeline_stage ... "build" ...
and+ () = record_pipeline_stage ... "run" ...
and+ () = match Current.state output with ... Error ... -> ... record_stage_failure ...
and+ out = output
in
out
```
This would allow to remove the star in `let* p = pipeline ...` below at line 137

(It absolutely works as is! But the less `let*` we have, the more predictable the pipeline is in the graphviz... ocaml-ci uses only `let+` and we are very close to having none too!)

punchagan: Thanks for the review and explanation, @art-w ! I've pushed an update changing this.

art-w:
Looks Great To Me! I don't know why ocaml-ci breaks so often with alpine :/

I added https://github.com/ocurrent/current-bench/issues/265 to remember that we should improve the readability of json errors!

PR #259, opened 2021-12-20
Add support for multi-value datapoints averaged by the benchmark

### LINKED ISSUE #132: multi-value datapoints averaged by the benchmark itself
In addition to `number` and `number array`, a metric should accept the `min/max/avg` triplet to achieve the same display result as `number array`. It could take the form of a json object: `{min: number; max: number; avg: number}`

My use case is that I already have computed the min/max/avg for variables that have thousands (if not millions) of points.

punchagan:
Currently, we only support single-value datapoints, or multi-value datapoints
as a list. For a multi-value datapoint, we compute the stats (min, max, and
avg) in the frontend and display the datapoint in the graph. This commit adds
support for multi-value datapoints where these stats have been calculated by
the benchmark itself. The value can be passed as a json object: `{min: number;
max: number; avg: number}`

Closes #132


gs0510 (approved):
Looks great! Thanks @punchagan!!

PR #262, opened 2021-12-21
Ensure that the key "avg" is present in values which are stats

punchagan:
6b792b9c0c080a7d2145051ce3631becbb348a25 added support for metrics which have
been averaged by the benchmarks themselves. This commit ensures that the "avg"
key is present in the averaged benchmarks. "min" and "max" are the other two
keys which would be nice to have, but don't break the UI if missing.


gs0510 (approved):
Ah nice catch, thanks @punchagan!!

PR #264, opened 2021-12-22
Fix VITE_CURRENT_BENCH_VERSION envvar in production

punchagan:
db19b3abba1ebbb1f3bda3944e3db021164c7756 added display of current-bench version
in the UI using an environment variable. But, the setup for production was
incorrect, and this commit fixes it.


gs0510:
aah yes, i saw this and was going to look into this!! thanks @punchagan!

gs0510 (approved):

PR #268, opened 2021-12-28
Fix frontend: add worker and docker image to url

art-w:
The worker and docker image choice was not reflected in the url. This causes issues when refreshing the page as they are reset to the first available.


gs0510 (commented):
Looks good to me! Thanks @art-w!!

The only thing I was thinking is that we redirect the user to `autumn.ocamllabs.io/owner/repo/pull_number` from the github check, so is it okay if we still do that?

art-w:
Yes it should still work! On arrival, the url will change to select the first available worker/image by adding `?worker=...&image=...` at the end of the url. In the future we might do something smarter when no worker/image are selected in the url (for example with graph overlays), so we shoud keep the github url as is!

punchagan (approved):
Looks good to me too!

PR #269, opened 2021-12-29
Add missing size units to the sizeRegex

punchagan:
Previously some of the size units were missing from the sizeRegex, which are
included in the sizeUnits array. This commit adds those units to the
sizeRegex. The units are explicitly listed down in the regex, because %re can
only be called on literal strings.


gs0510 (approved):
thanks! good catch @punchagan  :)

PR #270, opened 2022-01-03
Basic Prometheus and Alertmanager config

punchagan:
This PR adds basic configuration to get alertmanager and prometheus running, and adds a basic alert to verify that prometheus is able to connect to alertmanager. 

The Slack webhook api URL needs to be configured, and I don't have the permissions to add a new incoming webhook in our Slack instance. I'd need help with that. 

The expectation is to run node exporters on each of the nodes/cluster workers, and add alerts for some basic metrics. I'll add the scripts to get this running in a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,3 +104,7 @@ start-prometheus-alertmanager:
 stop-prometheus-alertmanager:
 	cd ./prometheus/ && \
 	docker-compose --env-file=../environments/production.env down
+
+.PHONY: start-node-exporter
+start-node-exporter:
+	./prometheus/scripts/start-node-exporter.sh --web.listen-address 0.0.0.0:10080
--- a/Makefile
+++ b/Makefile
@@ -94,3 +94,13 @@ rebuild-pipeline: ./local-test-repo/.git
 .PHONY: bench
 bench:
 	@cd ./local-test-repo/ && make -s bench
+
+.PHONY: start-prometheus-alertmanager
+start-prometheus-alertmanager:
+	cd ./prometheus/ && \
+	docker-compose --env-file=../environments/production.env up --detach
+
+.PHONY: stop-prometheus-alertmanager
+stop-prometheus-alertmanager:
+	cd ./prometheus/ && \
+	docker-compose --env-file=../environments/production.env down
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -11,6 +11,10 @@ scrape_configs:
     scrape_interval: 5s
     static_configs:
       - targets: ['localhost:9090', 'alertmanager:9093']
+  - job_name: 'node'
+    scrape_interval: 5s
+    static_configs:
+      - targets: [ $PROMETHEUS_SCRAPE_TARGETS ]
 
 alerting:
   alertmanagers:
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -1,0 +1,22 @@
+global:
+  scrape_interval: 5s # By default, scrape targets every 15 seconds.
+
+  # Attach these labels to any time series or alerts when communicating with
+  # external systems (federation, remote storage, Alertmanager).
+  external_labels:
+    monitor: 'current-bench-monitor'
+
+scrape_configs:
+  - job_name: 'prometheus'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:9090', 'alertmanager:9093']
+
+alerting:
+  alertmanagers:
+    - scheme: http
+      static_configs:
+        - targets: [ 'alertmanager:9093' ]
+
+rule_files:
+  - /etc/prometheus/alerts.yml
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -94,3 +94,17 @@ rebuild-pipeline: ./local-test-repo/.git
 .PHONY: bench
 bench:
 	@cd ./local-test-repo/ && make -s bench
+
+.PHONY: start-prometheus-alertmanager
+start-prometheus-alertmanager:
+	cd ./prometheus/ && \
+	(test -f ../environments/production.env && \
+	 docker-compose --env-file=../environments/production.env up --detach) || \
```
gs0510, line 102: perhaps we can only have it running for production? we don't really care about having it run for dev.

prometheus/prometheus.yml
```diff
@@ -0,0 +1,26 @@
...
+
+scrape_configs:
+  - job_name: 'prometheus'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:9090', 'alertmanager:9093']
+  - job_name: 'node'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:10080']
```
gs0510, line 17: What does this do?

prometheus/scripts/test-alerts.sh
```diff
@@ -0,0 +1,20 @@
+#! /usr/bin/env sh
+
+URL="http://localhost:9093/api/v1/alerts"
```
gs0510, line 3: We don't seem to use this anywhere, is this just for locally testing to see if the alert instance is up?

prometheus/scripts/start-alertmanager.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# Script to start alert manager.
+
+# Alertmanager doesn't provide a way to replace config variables, and we need
+# to write our own script to do this. See
+# https://github.com/prometheus/alertmanager/issues/504
+
+sed "s|\$ALERTMANAGER_SLACK_API_URL|$ALERTMANAGER_SLACK_API_URL|g" /config/alertmanager.yml > /tmp/alertmanager.yml  # Permissions to write to the same file is a problem, so we write a /tmp file
```
gs0510, line 8: Is the `/config/alertmanager.yml` path in the default docker image? I don't see it mounted anywhere, hence my confusion.

gs0510 (commented):
Great work @punchagan!! I have a few questions :))

1. For now, this only tells us that the instance is down right?
2. If the node is running inside the docker, can it have the permissions to look into resource usage?

Also, a bunch a questions on the review :))

prometheus/prometheus.yml
punchagan: This is configuration to scrape a `node_exporter` that I was using locally, but I'll remove this and add it along with the node_exporter script, etc.

prometheus/scripts/test-alerts.sh
punchagan: Yes, that's correct. I'll get rid of this script.

prometheus/scripts/start-alertmanager.sh
punchagan: It's mounted in the `docker-compose.yml` file.

punchagan:
> 1. For now, this only tells us that the instance is down right?

Currently, it doesn't even tell us that. It only tells us, if the prometheus instance is not able to talk to the alertmanager. 

>     2. If the node is running inside the docker, can it have the permissions to look into resource usage?

The plan is to run a `node_exporter` in each of the current-bench cluster workers, which will give information about the resources being used. This PR doesn't have that configuration or script. I was planning to send that as a separate PR, but I guess I could include it in the same PR.

prometheus/prometheus.yml
```diff
@@ -11,6 +11,13 @@ scrape_configs:
     scrape_interval: 5s
     static_configs:
       - targets: ['localhost:9090', 'alertmanager:9093']
+  - job_name: 'node'
+    scrape_interval: 5s
+    static_configs:
+      - targets:
+          - 'autumn.ocamllabs.io:10080'
```
punchagan, line 18: May be we should scrape this from the `production.env` config?

prometheus/scripts/start-alertmanager.sh
gs0510: Ah yes! i see it :)) thanks!

prometheus/prometheus.yml
gs0510: Ah that is indeed a good idea! Also, `comanche` is not our machine, so we can get rid of that! 😿

punchagan: I've pushed an update to generate this from our cluster pools configuration,

punchagan:
> Ah okay, this makes sense! If you think it's a good idea to include in this PR, that works! Else, this PR is also good to go for me :))

I pushed an update on this PR to include the script to run node_exporter, and to get the prometheus server to scrape these node exporters. I haven't yet added any alerts for the node resources getting depleted or node going down. I'll add those in a separate PR.

gs0510:
Great, thanks!

gs0510 (approved):

PR #271, opened 2022-01-04
Add prometheus alert when host is out disk space

punchagan:
When free space on disk is under 90%, send an alert.


gs0510 (commented):
LGTM, thanks @punchagan!

PR #272, opened 2022-01-06
Ignore NaNs when computing the mean

punchagan:
Previously, the mean included the count for the number of points, but ignored
NaN values. This seemed to make the mean value meaningless. This commit changes
the computation to completely ignore the mean values.


gs0510 (approved):
Nice catch, thanks @punchagan!!

PR #275, opened 2022-01-07
Frontend fix: display warning and old metrics rather than an empty page

art-w:
Fix for @punchagan https://github.com/ocurrent/current-bench/pull/267#discussion_r776245042 : When we don't have metrics for the latest run (because it was cancelled, failed, or still running), we display an orange warning and show the previous metrics.


punchagan:
Looks good, thanks!

punchagan (approved):

PR #276, opened 2022-01-07
Production fix: validate-env for graphql url

art-w:
In production, we use a local path `/_hasura/v1/graphql` to access the hasura graphql server via https (or http).


gs0510:
thanks @art-w!

PR #278, opened 2022-01-10
Fix for production

art-w:
- The frontend nginx running inside docker was configured to run on port 80, but the port is already used outside of docker
- The ocluster takes a long time to build because the opam dependencies are not cached by docker


gs0510 (approved):
Thanks @art-w !

PR #281, opened 2022-01-17
Document how to add new workers

art-w:
Should I expand on some of the steps or does it roughly makes sense?


gs0510:
This looks good to me! I worked through the doc and had a local worker just for testing :)

gs0510:
@shakthimaan can you take a look and see if there's something you think is missing? thanks!

gs0510 (approved):

HACKING.md
```diff
@@ -301,3 +301,79 @@ ERROR: for current-bench_db_1  Cannot start service db: driver failed programmin
...
+    --name=my-worker \
+    --state-dir=/path/to/internal/state \
+    --ocluster-pool=./pool-myworker.cap \
+    --docker-cpu=0-1,2-3
+```
+
+- The worker `--name` will show up in the build logs as `Building on <my-worker>` (for debugging)
+- The `--state-dir` should point to a directory where the worker will be able to store its data
+- The `--ocluster-pool` should be the filename created by the cluster.
+- Finally, the `--docker-cpu` is a comma separated list of CPUs available to the worker. If there are multiple choices `A,B,C` then the worker will be able to run as many benchmarks at the same time. If there are no comma, then only one benchmark will be run at a time. A range `A-B` allow for multicore benchmarks on the cpus A to B, while a single `X` cpu number is singlecore. If ranges are used, they should have the same number of CPUs and not overlap!
```
shakthimaan, line 379: `... CPU number is for a single core`, perhaps?

gs0510:
Thanks @shakthimaan!

PR #283, opened 2022-01-20
Scrape metrics from the pipeline server

punchagan:
The ocurrent routes already configure /metrics to expose some ocaml gc metrics
and some basic metrics like github webhook stats, cache stats, etc.  This
commit configures Prometheus to scrape these metrics.


gs0510:
Looks great, thanks!

PR #284, opened 2022-01-28
Couple of fixes with handling multiple benchmarks data

punchagan:
This PR adds a couple of fixes for handling data of benchmark runs with multiple benchmarks. We support having a list of benchmarks, when each of them should be named.

1. Hide the benchmarks menu only when unnamed benchmarks are used. 
2. Allow having test-names repeating across different benchmarks


art-w:
Good catches! Thanks :)

PR #287, opened 2022-01-28
HTTP API end-point for capturing benchmark results

punchagan:
TODO:

-  [x] Add mechanism to authenticate users

### DIFF
--- a/environments/development.conf
+++ b/environments/development.conf
@@ -1,10 +1,12 @@
-[
-  {
-    "name": "local/local",
-    "image": "ocaml/opam:debian-11-ocaml-4.14"
-  },
-  {
-    "name": "local/local",
-    "image": "ocaml/opam:debian-11-ocaml-4.13"
-  }
-]
+{
+  "repositories": [
+    {
+      "name": "local/local",
+      "image": "ocaml/opam:debian-11-ocaml-4.14"
+    },
+    {
+      "name": "local/local",
+      "image": "ocaml/opam:debian-11-ocaml-4.13"
+    }
+  ]
+}
--- a/environments/development.conf
+++ b/environments/development.conf
@@ -8,5 +8,11 @@
       "name": "local/local",
       "image": "ocaml/opam:debian-11-ocaml-4.13"
     }
+  ],
+  "api_tokens": [
+      {
+      "repo": "local/local",
+      "token": "super-secret-token"
+    }
   ]
 }
--- a/pipeline/lib/api.ml
+++ b/pipeline/lib/api.ml
@@ -1,12 +1,49 @@
 open Lwt.Infix
 module Utils = Cohttp_lwt_unix
 
+let request_token req =
+  let headers = Cohttp.Request.headers req in
+  let token = Cohttp.Header.get headers "Authorization" in
+  let open String in
+  match token with
+  | Some token ->
+      if starts_with ~prefix:"Bearer " token
+      then
+        let n = length token in
+        let m = length "Bearer " in
+        sub token m (n - m)
+      else Fmt.failwith "Invalid token"
+  | _ -> ""
+
+let authenticate_token req api_tokens =
+  (* This function only that the token is a valid token, and doesn't process
+     the body of the request. Essentially, we only do "authentication" of the
+     token. The authorization is done in a separate function.*)
+  (match api_tokens with
+  | [] -> failwith "Server configuration error: API tokens not configured."
+  | _ -> ());
+
+  let token = request_token req in
+  match token with
+  | "" -> failwith "Invalid token"
+  | _ ->
+      List.find_opt
+        (fun (t : Config.api_token) -> String.equal t.token token)
+        api_tokens
+
+let authorize_token (token : Config.api_token option) repository =
+  let repo_id = Repository.info repository in
+  match token with
+  | Some token ->
+      if String.equal token.repo repo_id then () else failwith "Invalid token"
+  | _ -> failwith "Invalid token"
+
 let check_key key json =
   match Yojson.Safe.Util.(member key json) with
   | `Null -> Fmt.failwith "Key '%s' is missing" key
   | s -> s
 
-let make_benchmark_from_request ~conninfo ~body =
+let make_benchmark_from_request ~conninfo ~body ~token =
   let json_body = body |> Yojson.Safe.from_string in
   let open Yojson.Safe.Util in
   let repository_name = check_key "repo_name" json_body |> to_string in
@@ -48,6 +85,7 @@ let make_benchmark_from_request ~conninfo ~body =
     Models.Benchmark.make ~duration ~run_at ~repository ~worker ~docker_image
       ?build_job_id ?run_job_id
   in
+  authorize_token token repository;
   let serial_id =
     Storage.setup_metadata ~repository ~conninfo ~worker ~docker_image
   in
@@ -58,15 +96,16 @@ let make_benchmark_from_request ~conninfo ~body =
              [ Current_bench_json.of_json bench ]));
   Storage.record_success ~conninfo ~serial_id
 
-let capture_metrics conninfo =
+let capture_metrics conninfo api_tokens =
   object
     inherit Current_web.Resource.t
 
-    method! post_raw _ _ body =
+    method! post_raw _ req body =
       Cohttp_lwt.Body.to_string body >>= fun body ->
       let status, response =
         try
-          make_benchmark_from_request ~conninfo ~body;
+          let token = authenticate_token req api_tokens in
+          make_benchmark_from_request ~conninfo ~body ~token;
           let status = `OK in
           let response = `Assoc [ ("success", `Bool true) ] in
           (status, response)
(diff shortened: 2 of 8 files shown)


pipeline/lib/api.ml
```diff
@@ -0,0 +1,98 @@
...
+      ?build_job_id ?run_job_id
+  in
+  { benchmark; benchmarks; repository; worker; docker_image }
+
+let capture_metrics conninfo =
+  object
+    inherit Current_web.Resource.t
+
+    method! post_raw _ _ body =
+      (* FIXME: Validate that required fields are present in the body!     *)
```
gs0510, line 71: Is this validation for the benchmark data? (Like we do for other benchmarks?)

gs0510 (commented):
Awesome work @punchagan! I think the missing bits are 1. Benchmark validation and 2. User validation. We should only allow folks that we permit to access the API. :)

pipeline/lib/api.ml
punchagan: No, this is validation for the data in the API. The validation of the benchmark data is already happening, since I use the `Current_bench_json.of_json` function on the benchmark data submitted when saving it to the DB. This comment was an outdated one, since I also added the request validation after initially adding this FIXME as a reminder to myself.

punchagan:
> Awesome work @punchagan! I think the missing bits are 1. Benchmark validation and 2. User validation. We should only allow folks that we permit to access the API. :)

Thanks @gs0510. 

I posted a comment on 1. in the inline comment - Benchmark validation should already happening in the calls to `Current_bench_json.of_json`. 

I pushed a commit to add user validation based on some inputs from @art-w on how to do it. The validation is very simplistic in that we allow anyone with a single token access to the API (since we have only one consumer for the end-point).

art-w:
Ah sorry for the confusion, I meant that each user would have its own token / password. Here it seems that we would have a single password, which doesn't guarantee that we can't push to the wrong repository?

environments/development.conf
```diff
@@ -1,10 +1,13 @@
...
+    {
+      "name": "local/local",
+      "image": "ocaml/opam:debian-11-ocaml-4.14"
+    },
+    {
+      "name": "local/local",
+      "image": "ocaml/opam:debian-11-ocaml-4.13"
+    }
+  ],
+  "api_token": "super_secret_token"
```
art-w, line 12: This could be a list similar to `repositories`:

```json
"apis": [
  { "name": "local/local", "token": "my secret" },
  ...
]
```

punchagan:
> Ah sorry for the confusion, I meant that each user would have its own token / password. Here it seems that we would have a single password, which doesn't guarantee that we can't push to the wrong repository?

Ah, sorry. I missed (or forgot about) the clear spec with types in your original message (on Slack). I'll update this PR!

pipeline/lib/api.ml
```diff
@@ -0,0 +1,113 @@
...
+          in
+          let serial_id =
+            Storage.setup_metadata ~repository ~conninfo ~worker ~docker_image
+          in
+          ignore
+            (benchmarks
+            |> List.map (fun bench ->
+                   Json_stream.db_save ~conninfo benchmark
+                     [ Current_bench_json.of_json bench ]));
+          Storage.record_success ~conninfo ~serial_id;
```
art-w, line 101: This can probably be moved inside `make_benchmark_from_request` by passing it `~conninfo`. By doing so, the record `Api.t` could be removed and I think the code would be simpler (?)

pipeline/lib/api.ml
```diff
@@ -0,0 +1,113 @@
+open Lwt.Infix
+module Utils = Cohttp_lwt_unix
+
+type t = {
+  benchmark :
+    version:int ->
+    benchmark_name:string option ->
+    test_index:int ->
+    Yojson.Safe.t ->
+    Models.Benchmark.t;
```
art-w, line 10: Just to clarify why I'm suggesting to remove this record `t`: in ocaml it's very rare that you need a function inside a record... so you really have to justify it when you do it ^^

When reading the code, this type sends the message "I'm going to arbitrarily set this `benchmark` field with a lot of different unknown functions" (which isn't a good sign!). But here it's just an intermediate value :)

pipeline/lib/api.ml
art-w: Github did not remember the line range I selected! It was 93..101 (so all the code after `make_benchmark_from_request`)

pipeline/lib/api.ml
```diff
@@ -0,0 +1,113 @@
...
+    match d with Some (Some duration) -> duration | _ -> Ptime.Span.of_int_s 0
+  in
+  let benchmarks = check_key "benchmarks" json_body |> to_list in
+  let build_job_id = Some "" in
+  let run_job_id = Some "" in
+  let worker = "remote" in
+  let docker_image = "external" in
+  let commit =
+    let open Current_git.Commit_id in
+    v ~repo:"git://pipeline/" ~gref:"HEAD" ~hash:commit
```
art-w, line 54: It's a bit sad that we need to do that to create a repository! :D

pipeline/lib/api.ml
punchagan: Thanks for clarifying!

punchagan:
Thanks for your review, @art-w! I've force pushed changes to incorporate the changes suggested by you. Could you take a look again, please?

pipeline/lib/api.ml
```diff
@@ -0,0 +1,120 @@
...
+          let token = authenticate_token req api_tokens in
+          make_benchmark_from_request ~conninfo ~body ~token;
+          let status = `OK in
+          let response = {|{"success": true}|} in
+          (status, response)
+        with
+        | Yojson.Json_error e | Yojson.Safe.Util.Type_error (e, _) | Failure e
+        ->
+          ( `Bad_request,
+            Printf.sprintf "{\"success\": false, \"error\": \"%s\"}" e )
```
art-w, line 116: Useful trick: `printf` has `%S` (in uppercase) which produces a string with surrounding quotes and internal quotes are escaped with `\` (so it's better to have `%S` than `\"%s\"`)

```ocaml
# Printf.sprintf "%S" {|hello "world" !|} ;;
- : string = "\"hello \\\"world\\\" !\""
```

(but it would even safer if `response` was produced by `Yojson.Safe.to_string`)

art-w:
It looks good! Thanks for fixing the token authorization :)

pipeline/lib/api.ml
punchagan: Oh, thanks! I wasn't aware of this. :)

pipeline/lib/api.ml
```diff
@@ -0,0 +1,120 @@
...
+          let token = authenticate_token req api_tokens in
+          make_benchmark_from_request ~conninfo ~body ~token;
+          let status = `OK in
+          let response = {|{"success": true}|} in
+          (status, response)
+        with
+        | Yojson.Json_error e | Yojson.Safe.Util.Type_error (e, _) | Failure e
+        ->
+          ( `Bad_request,
+            Printf.sprintf "{\"success\": false, \"error\": \"%S\"}" e )
```
art-w, line 116: Ah the `%S` already adds surrounding quotes to the string, no need to wrap it in `\"...\"` ^^

pipeline/lib/api.ml
punchagan: I wasn't aware of `%S`, thanks!
 
I've updated the PR to produce the response using `Yojson.Safe.to_string`.  Appreciate your inputs! :)

pipeline/lib/api.ml
```diff
@@ -0,0 +1,125 @@
...
+        with
+        | Yojson.Json_error e | Yojson.Safe.Util.Type_error (e, _) | Failure e
+        ->
+          let response =
+            `Assoc [ ("success", `Bool true); ("error", `String e) ]
+            |> Yojson.Safe.to_string
+          in
+          (`Bad_request, response)
+      in
+      let body = response |> Cohttp_lwt.Body.of_string in
```
art-w, line 123: Perhaps the call to `Yojson.Safe.to_string` can be factored here once?

pipeline/lib/api.ml
```diff
@@ -0,0 +1,125 @@
...
+          let status = `OK in
+          let response =
+            `Assoc [ ("success", `Bool true) ] |> Yojson.Safe.to_string
+          in
+          (status, response)
+        with
+        | Yojson.Json_error e | Yojson.Safe.Util.Type_error (e, _) | Failure e
+        ->
+          let response =
+            `Assoc [ ("success", `Bool true); ("error", `String e) ]
```
art-w, line 118: Probably not a "true" success? :D

punchagan: Ah, good catch. I should've wrapped up my day. I'll update this!

pipeline/lib/api.ml
punchagan: Yes!

pipeline/lib/api.ml
art-w: No worries! Sorry for flooding you with lots of small comments this late, I'm very happy with the PR!

pipeline/lib/api.ml
punchagan: Ah, I see. :+1:

art-w:
This is so cool @punchagan ! Will merge as soon as ocaml-ci terminates :)

PR #288, opened 2022-01-28
nginx frontend: fix for urls containing slashes

art-w:
In production, the `/` was causing some 404 not found... I believe that `/index.html$is_args$args` was expanding to `/index.html?foo=bar/blah` with the second `/` interpreted as a directory separator?

In any case, we don't need to pass arguments to the static `index.html`.


gs0510:
Thanks @art-w!

PR #289, opened 2022-01-28
Fix url missing pull number

art-w:
When we redirect to add the worker / docker image, we loose the PR number!


gs0510:
thanks a ton @art-w!! :)) thank you for finding where the pull number was disappearing also!  <3

PR #292, opened 2022-02-01
Return better HTTP error codes for API requests

punchagan:
This might be an over-kill for our simple one consumer end-point, but I also was trying to get a feel for using exceptions in OCaml.

### DIFF
--- a/pipeline/lib/api.ml
+++ b/pipeline/lib/api.ml
@@ -1,6 +1,16 @@
 open Lwt.Infix
 module Utils = Cohttp_lwt_unix
 
+exception Invalid_token
+
+exception Missing_token
+
+exception Unauthorized
+
+exception Server_config_error
+
+exception Data_validation_error of string
+
 let request_token req =
   let headers = Cohttp.Request.headers req in
   let token = Cohttp.Header.get headers "Authorization" in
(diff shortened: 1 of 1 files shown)


pipeline/lib/api.ml
```diff
@@ -111,11 +121,26 @@ let capture_metrics conninfo api_tokens =
...
+            in
+            (`Unauthorized, response)
+        | PermissionError e ->
+            let response =
+              `Assoc [ ("success", `Bool false); ("error", `String e) ]
+            in
+            (`Forbidden, response)
+        | ServerError e ->
+            let response =
+              `Assoc [ ("success", `Bool false); ("error", `String e) ]
```
art-w, line 141: Perhaps this could be shorter with:

```ocaml
let error_response msg = `Assoc [ ... ]
```

pipeline/lib/api.ml
```diff
@@ -1,6 +1,12 @@
 open Lwt.Infix
 module Utils = Cohttp_lwt_unix
 
+exception AuthError of string
+
+exception PermissionError of string
+
+exception ServerError of string
```
art-w, line 8: This is a bit subjective and you'll have to choose what you prefer, but it can be helpful to define more domain specific exceptions. Here you are in a middle ground between the HTTP errors and the domain... and so you need `of string` (which to the future reader means "I will reuse this exception for different cause of failures")

Defining exceptions acts as a warning "here is what can go wrong":
```ocaml
exception Api_not_enabled
exception Invalid_token
exception Wrong_token
```

This avoids the redundant `AuthError "Invalid token"` and mean all the user facing error messages can be defined in one place, independently of the logic.

But in this specific situation, you could also go all in with the HTTP errors in order to simplify the unique `try...with`:

```ocaml
exception Error of [`Bad_request | `Unauthorized | ... ] * string
let invalid_token () = raise (Error (`Unauthorized, "Invalid token"))
let wrong_token () = raise (Error ...)
```

(I like this less as it makes it harder to add unit tests on the validation, but it's less verbose and achieves the same result)

pipeline/lib/api.ml
```diff
@@ -111,11 +121,26 @@ let capture_metrics conninfo api_tokens =
...
+        | PermissionError e ->
+            let response =
+              `Assoc [ ("success", `Bool false); ("error", `String e) ]
+            in
+            (`Forbidden, response)
+        | ServerError e ->
+            let response =
+              `Assoc [ ("success", `Bool false); ("error", `String e) ]
+            in
+            (`Internal_server_error, response)
```
art-w, line 143: I don't know if we get "nice" error messages in the log if an unanticipated exception is raised? (and not catched by this `try`)
- user should get an "internal server error"
- we should get a dump of the error and backtrace in the log to investigate :)

It's very likely that ocurrent/cohttp already handle this! Otherwise look into adding a "catchall" `try ... with ... | error -> ...` and [Printexc](https://ocaml.org/api/Printexc.html) :)

punchagan: Yes, ocurrent/cohttp already handle this!

pipeline/lib/api.ml
punchagan: Yep, thanks!

pipeline/lib/api.ml
punchagan: Thanks for the clear explanation. Domain specific exceptions make sense, thanks!

art-w:
Thanks!

PR #293, opened 2022-02-01
Make log links noticeable by removing log ID from link text

### LINKED ISSUE #291: Make logs link more noticeable
From #285,

>Thanks I didn't notice this link, I think it could be made more noticeable (e.g. using the same style blue+underline for the link, and personally I think that the name of the build "110849-ocluster-build-87cd3f" only adds some "noise" and is not very useful for the end user, the link could just be named "execution logs").

Right now there;s ids in the logs that doesn't help the user figure out that the execution logs is indeed something useful. Perhaps we can add just a `Logs` link and hide the uuids?

punchagan:
Closes #291

### DIFF
--- a/frontend/src/CommitInfo.res
+++ b/frontend/src/CommitInfo.res
@@ -70,8 +70,9 @@ let renderExternalLink = (~style=linkStyle, ~href, text) => {
 let url: string = %raw(`import.meta.env.VITE_OCAML_BENCH_PIPELINE_URL`)
 
 let renderJobIdLink = (jobId, ~text) => {
+  let style = [Sx.text.xs]
   let href = url ++ "/job/" ++ jobId
-  renderExternalLink(~href, text)
+  renderExternalLink(~style, ~href, text)
 }
 
 let renderCommitLink = (~style=linkStyle, repoId, commit) =>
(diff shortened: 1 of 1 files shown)


art-w:
Nice! The wording will definitively help, but perhaps part of the confusion is also that we use unstyled black links with no underline? (so it's not obvious what is clickable or not)

punchagan:
On hover, we do change color. So, I feel it may be sufficient with the new wording. 

With external links shown in the primary color, it looks like this.  Does this look alright - I'm not sure? Or do you have any specific ideas on how to display these links? @gs0510 thoughts?

image

punchagan:
Also, I feel we can move the link for Execution logs somewhere else. The other 3 items - last commit, environment and build status are giving information about the current build. The execution logs link is just a link and shouldn't show up along with other "info items", I think.

gs0510:
I agree that we can move the `View Logs` somewhere else. Maybe the status and the `view logs` can be bunched together (and it's more clear that way that one can look at the logs next to the status)? I don't have other ideas at the moment though. :)

punchagan:
What do you think about this? 

image

Older builds with different build and execution logs
image

gs0510:
Looks great imo!

punchagan:
Merging after :+1: from @art-w and @gs0510

PR #294, opened 2022-02-01
Add unit tests

art-w:
This is a WIP to add unit tests to parts of the pipeline. In its current state it's mostly a prototype to ease adding more tests in the future as I believe it would help a lot:

- I started by adding some unit tests for the new `api` since the code was still fresh in my mind. I discovered that I had missed during the review that tokens should be globally unique, which might cause some confusion if we treat tokens as passwords (?) Also I think that the unit tests could be simpler if the validation logic was not so dependent on Cohttp. @punchagan > Do you mind taking a look? :)

- Then I added some tests to the `Json_stream` parser since I already know it's not perfect... but this revealed an unexpected bug when the log output starts directly with a json. I plan to add a lot more tests to this module now!

- The main issue for writing unit tests is that we perform a lot of side effects (via ocurrent and postgresql), so I added a "mock" Postgresql module which allow the unit tests to simulate the sql server by specifying the expected queries and their outcome. I'm not yet satisfied with the sql "string" validation, so I'm open to any ideas!

In order to run the tests, you have to be inside the `pipeline/` directory and execute `dune runtest` or `dune exec tests/test.exe`. @gs0510 > Do you know if this something we could automatically run on new PR?

The overall idea is that the `tests/dune` imports the code from `pipeline/*.ml` so that the modules `tests/*_test.ml` can access the internal functions (that would otherwise be painful to expose in the `pipeline` library). This also allow us to "override" a library to mock it (see `tests/postgresql.ml` and the missing dependency to `postgresql` in `tests/dune`). I think we should be able to do something similar for modules that depend on ocurrent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,3 +101,11 @@ stop-prometheus-alertmanager:
 .PHONY: start-node-exporter
 start-node-exporter:
 	./prometheus/scripts/start-node-exporter.sh --web.listen-address 0.0.0.0:10080
+
+.PHONY: runtest
+runtest: ./local-test-repo/.git
+	docker-compose \
+		--project-name="current-bench" \
+		--file=./environments/development.docker-compose.yaml \
+		--env-file=./environments/development.env \
+		exec pipeline bash -c 'cd /mnt/project; opam exec -- dune runtest'
--- a/pipeline/tests/api_test.ml
+++ b/pipeline/tests/api_test.ml
@@ -1,6 +1,6 @@
 let example_conf =
   [
-    { Config.repo = "myowner/repo"; token = "p455word!" };
+    { Config.repo = "myowner/myrepo"; token = "p455word!" };
     { Config.repo = "other/fst"; token = "token-should-be-unique" };
     { Config.repo = "other/snd"; token = "token-should-be-unique" };
   ]
@@ -16,7 +16,7 @@ let cohttp_unauth_request =
   Cohttp.Request.make uri
 
 let server_not_configured =
-  Alcotest.test_case "server not configured" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "server not configured" `Quick @@ fun () ->
   let req = cohttp_request ~bearer:"Bearer 123" in
   try
     let _ = Api.authenticate_token req [] in
@@ -32,7 +32,7 @@ let missing_bearer =
   with Api.Missing_token -> ()
 
 let invalid_bearer =
-  Alcotest.test_case "invalid bearer" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "invalid bearer" `Quick @@ fun () ->
   let req = cohttp_request ~bearer:"Missing Bearer Prefix" in
   try
     let _ = Api.authenticate_token req example_conf in
@@ -40,7 +40,7 @@ let invalid_bearer =
   with Api.Invalid_token -> ()
 
 let wrong_token =
-  Alcotest.test_case "wrong token" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "wrong token" `Quick @@ fun () ->
   let req = cohttp_request ~bearer:"Bearer wrong-token" in
   match Api.authenticate_token req example_conf with
   | None -> ()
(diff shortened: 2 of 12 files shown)


punchagan:
Adding unit tests is a great idea! I was wondering about this today, while going through the `json_stream` code and wanting to tweak it and play around with it. I'll take a look at the tests tomorrow. :)

pipeline/tests/api_test.ml
```diff
@@ -0,0 +1,139 @@
+let example_conf =
+  [
+    { Config.repo = "myowner/myrepo"; token = "p455word!" };
+    { Config.repo = "other/fst"; token = "token-should-be-unique" };
+    { Config.repo = "other/snd"; token = "token-should-be-unique" };
+  ]
+
+(* TODO: Cohttp should not be required for testing auth *)
```
gs0510, line 8: Why should `Cohttp` not be used? Because we want to send dummy requests instead? :)

pipeline/tests/dune
```diff
@@ -0,0 +1,11 @@
...
+ (name test)
+ (libraries alcotest alcotest-lwt current current.fs current_docker
+   current_git current_github current_ocluster capnp-rpc-unix current_web
+   dockerfile fmt.tty logs logs.fmt prometheus rresult uri yojson str)
+ (preprocess
+  (pps ppx_deriving_yojson)))
+
+; copy everything but *.pp.ml
+
+(copy_files ../lib/**[!.]??.ml)
```
gs0510, line 11: Why do we need to copy these files? Can we just have public modules instead?

gs0510 (commented):
This is super cool @art-w! I have some questions, and I'll give it a pass again tomorrow for new inspiration. :)

pipeline/tests/api_test.ml
punchagan: I guess so. I think we could refactor the validation code to use the token string directly instead of the request to make the testing easier.

pipeline/tests/dune
art-w: Yeah this copy is the "hacky" part. The proper way would be to expose all the modules and functions that we want to test in the `pipeline.mli` library, which is a bit tedious:

```ocaml
module Api : sig
  val authenticate_token : ...
  val authorize_token : ...
end
...
```

But the main issue is that we can't mock the postgresql module if we do that, so it becomes a lot harder to test the side effects part. By copying, we get "free" access to the internal functions for testing **and** we gain the ability to remove/simulate the side effectful libraries :)

(The copy is fairly benign as dune copies the files in the hidden `_build/` folder, not the `tests/` directory that you interact with)

pipeline/tests/api_test.ml
art-w: It's just a bit weird that we have to simulate a cohttp request for all of the tests. It makes sense for `invalid_bearer` (because it is testing an invalid header), but for example `find_repository` doesn't really need it, we only want to test the token validation here and not the extraction of the token from the headers. (it makes it a bit easier to understand a problem when just a few tests can independently fail)

punchagan:
We'd need to update the API tests slightly with #292 merged. I'd be happy to do that @art-w !

art-w:
I was going to do it to not bother you, but I would love if you could try the unit tests! I added you on github to the collaborators of my current-bench fork, so you should be able to push to this PR by adding commits on the `test` branch (let me know if it doesn't work, I'm not very familiar with this)

punchagan:
I have run the tests already in the current state, and they run well. I will update them to work with the latest changes on `main`. Also, I think if you enable the "[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" flag on this PR, I should be able to push to this branch without access to your fork. Could you edit the PR to enable that option?  If that doesn't work, I'll use your invitation to be a  collaborator on your fork.

art-w:
Hmm the checkbox "Allow edits from maintainers" was already checked, but I'm not sure if we have the required write access.

punchagan:
I'll check tomorrow, and confirm. If not, I'll use the invite you shared. :)

punchagan:
> I'll check tomorrow, and confirm. If not, I'll use the invite you shared. :)

I was able to push onto this branch, without being added as a collaborator on your fork. I'll try and update this PR with improved tests.

punchagan:
I've updated this PR to
- updated tests to work on the code in main
- updated opam file to install `alcotest` and `alcotest-lwt` to be able to run tests locally. It still doesn't help with running tests in ocaml-ci. Could you help with that, @art-w @gs0510 ? 
- added a make file target to run tests locally
- Removed `-t` flag on opam install in production docker container

punchagan:
> * updated opam file to install `alcotest` and `alcotest-lwt` to be able to run tests locally. It still doesn't help with running tests in ocaml-ci. Could you help with that, @art-w @gs0510 ?

I updated the pipeline.opam file before for running tests locally. Updating the `current-bench.opam` file fixed the missing deps issue on `ocaml-ci`! :tada:

Makefile
```diff
@@ -108,3 +108,11 @@ stop-prometheus-alertmanager:
 .PHONY: start-node-exporter
 start-node-exporter:
 	./prometheus/scripts/start-node-exporter.sh --web.listen-address 0.0.0.0:10080
+
+.PHONY: runtest
+runtest: ./local-test-repo/.git
```
gs0510, line 113: Why do we need this make target, when running `dune runtest` also works :)

gs0510 (commented):
Looks great! I just have one comment @punchagan

Makefile
punchagan: `dune runtest` would run the tests on the host machine. I'd like to run tests in the `pipeline` docker container and not rely on having a local opam setup. Does that make sense or am I missing some thing?

gs0510: Ah yes. I would install the dependencies for running the test in my local switch rather than using the make command. But maybe the other workflow can also be used by other people? I don't have strong opinions on this :)

punchagan: I guess it does help with editor support, etc., to have the local opam switch updated with all the dependencies. But, it could also seem like extra work to keep the local switch updated with all the deps. May be it's just that I'm still not super comfortable with OCaml and related tools, that I prefer running the tests in the docker container which is guaranteed to have all the required deps for a particular commit that I'm running tests on :)

punchagan:
Rebased this on the lastest `main` to see if tests pass

art-w:
Thanks! I added a version for cmdliner in `pipeline.opam` to fix the deployability failure :)

gs0510:
CI is passing 🎉 ! cc @art-w, @punchagan  :) I'll redeploy once this is also merged 😁

punchagan:
I think we can merge this, now? @art-w @gs0510 ? We can work on follow-ups (clean up of API functions/tests) in a separate PR, I think.

gs0510:
Yes, sounds good! :)

PR #295, opened 2022-02-03
Fix benchmark name unique constraint

punchagan:
c52c25a965a8a871f4c4a88f5894f2bb3ced4237 added a UNIQUE constraint
that includes the benchmark_name column. Postgres doesn't consider
NULL values to be equal and so the conflict on the UNIQUE constraint
with benchmark_name never gets triggered. 

This makes duplicate rows in the DB for the same metric, when we parse 
the JSON from the stream in multiple bits.

This PR has a couple of commits, one fixing the problem at the application level, and another to fix it at the DB level

### DIFF
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -340,7 +340,9 @@ let make = () => {
   switch route {
   | Error({reason}) => <ErrorView msg={reason} />
   | Ok(Main) => <RepoView worker={None} />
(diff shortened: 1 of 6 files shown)


gs0510:
LGTM ! Thanks @punchagan :)

art-w:
Good catch, thanks!

PR #296, opened 2022-02-04
Capture line numbers of where metrics appear in the logs

punchagan:
This is a WIP PR, which is almost close to being finished.

### DIFF
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -149,8 +149,8 @@ module V2 = struct
     | `Assoc vs ->
         let vs, units =
           let keys = List.map (fun (key, _) -> key) vs in
-          if not (List.mem "avg" keys)
-          then failwith "V2: Missing key *avg* in value";
+          if not (List.mem "avg" keys) then
+            failwith "V2: Missing key *avg* in value";
           List.split
           @@ List.map
                (fun (key, v) ->
@@ -191,8 +191,7 @@ module V2 = struct
     | _ ->
         failwith
         @@ "V2: trend should be lower-is-better, higher-is-better or not set. "
-        ^ trend
-        ^ " is not valid.");
+        ^ trend ^ " is not valid.");
     { name; description; value; units; trend }
 
   let metric_of_json_v1 (name, value) =
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -241,9 +241,7 @@ let version = 2
 let of_json json = Latest.of_json json
 
 let of_list jsons =
-  List.fold_left
-    (fun acc json -> Latest.merge acc [ Latest.of_json json ])
-    [] jsons
+  List.fold_left (fun acc json -> Latest.merge acc [ of_json json ]) [] jsons
 
 let to_list ts =
   List.map
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -24,6 +24,7 @@ module Json = struct
 
   let to_string_option = function `String s -> Some s | _ -> None
   let to_int_option = function `Int s -> Some s | _ -> None
+  let to_list_option = function `List xs -> Some xs | _ -> None
   let to_list = function `List xs -> xs | _ -> invalid_arg "Json: not a list"
 
   let to_assoc = function
@@ -67,12 +68,15 @@ module V2 = struct
     | Floats of float list
     | Assoc of (string * float) list
 
+  type line_range = int * int
+
   type metric = {
     name : string;
     description : string;
     value : value;
     units : string;
     trend : string;
+    lines : line_range list;
   }
 
   type result = { test_name : string; metrics : metric list }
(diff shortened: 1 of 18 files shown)


pipeline/lib/json_stream.ml
```diff
@@ -125,9 +174,10 @@ module Save = struct
     let run_job_id = Some worker_job_id in
     let json_stream = job_output_stream worker_job_id in
     Lwt_stream.fold
-      (fun jsons acc ->
-        let jsons = String.concat "\n" jsons in
-        let jsons = Util.parse_jsons jsons in
+      (fun (jsons, line_ranges) acc ->
+        let jsons = jsons |> List.rev |> List.map Yojson.Safe.from_string in
```
art-w, line 178: As our parser doesn't guarantee valid json (only well formed brackets of `{}[]""`), we should probably catch and filter out the values that Yojson rejects :)

pipeline/lib/json_stream.ml
```diff
@@ -38,30 +37,64 @@ let json_step stack chr =
...
+            | (start, _) :: positions ->
+                (json :: parsed, (start, pos) :: positions)
+            | [] -> (json :: parsed, positions))
+        | None, Some _start -> (
+            match positions with
+            | [] -> (parsed, (pos, None) :: positions)
+            | ((Some _, Some _) as hd) :: positions ->
+                (parsed, (pos, None) :: hd :: positions)
+            | (None, Some y) :: positions -> (parsed, (pos, Some y) :: positions)
+            | _ :: positions -> (parsed, (pos, None) :: positions))
```
art-w, line 93: The code could be clearer if the `positions` were stored in the same list as `parsed` as they are supposed to have the same length in the end... but it's harder to maintain this invariant when you have `string list * range list` rather than `(string * range) list`.

I would add a `start_line : int` to the `state` to remember when we first started adding stuff to the buffer, so that we can push `(state.start_line, state.lines, str_json)` to the `parsed` list when we are done. This would also avoid the pattern matching on `positions` to discover the current range start/stop :)

pipeline/lib/json_stream.ml
punchagan: Should we filter it out or let Yojson crash? It could be invalid non-JSON that our parser thinks is JSON or incorrect JSON generated by our users too, in which case having Yojson crash would be useful to debug builds where benchmarks don't show up.

art-w: Right... but it's more likely that our parser gets confused and selects "invalid json" that yojson will reject, than users producing "invalid json that's still recognized by our parser but not by yojson". As long as we still signal the schema errors in the json raised by `Current_bench_json`, it should be fine to treat the Yojson failures as "a mistake in our parser" (?)

punchagan: Yes, that makes sense. I guess users would use a valid JSON library to produce their JSON and the errors would be in the schema, if at all, rather than the structure of the JSON.

pipeline/lib/json_stream.ml
```diff
@@ -38,27 +42,38 @@ let json_step stack chr =
...
-        (Some str, make_json_parser ()))
+        let st = make_json_parser () in
+        ( Some str,
+          { st with lines = state.lines; start_line = state.start_line } ))
   | hd :: _ as stack ->
-      if hd <> '\n' then Buffer.add_char state.current chr;
-      (None, { state with stack })
+      if chr <> '\n' || hd = '"' then Buffer.add_char state.current chr;
+      let start_line =
+        if Buffer.length state.current == 1 then state.lines
```
art-w, line 64: It took me a bit of time to understand the `1` and not `0`, but that's because we `add_char` on the line above... Perhaps `if state.stack = []` is a simpler criteria independent of side effects? (meaning that the parser was not parsing before, but now is)

(small typo: it's a single `=` in caml, even if `==` works fine for integers)

pipeline/lib/current_bench_json.ml
```diff
@@ -69,18 +74,19 @@ module V2 = struct
     | Floats of float list
     | Assoc of (string * float) list
 
+  type line_range = int list option
```
art-w, line 77: Couple of questions because I expected `type line_range = int * int` or `{ start : int ; stop : int }`:
- Why is there an `option`?
- Can the list contains more or less than two elements?

punchagan: The `option` me using a `list` instead of a `int * int` was an artefact from the older code and me not thinking clearly.  I've now removed the option and `type line_range = int * int` as expected.

pipeline/lib/json_stream.ml
punchagan: This is not a change from this PR. So, if it seems reasonable to you, let's add this in a separate PR?

pipeline/lib/json_stream.ml
punchagan: Yes, checking the `stack` state before and now seems nicer, indeed! thanks

pipeline/lib/json_stream.ml
punchagan: Nevermind, I pushed a couple of commits on this branch to handle "incorrect" JSON parsed by our parser and rejected by Yojson.

punchagan:
@art-w can you take another look at this PR and merge it, if it looks good to you? This will at least start capturing the line numbers of metrics in the DB. The UI link will not be functional until ocurrent PR is merged, but we could hide the UI element, if required.

art-w (approved):

art-w:
LGTM Thanks!

Issue #297, opened 2022-02-09
Github status update


art-w:
There's an issue with the GitHub status update of the application: it automatically goes to the green "passed" state, even though the benchmarks are still running. I'm looking into it, but this seems to be related to the way we use `Current.state`.

It would also be nice if we could "check" that our behavior is still correct when we update the pipeline code!

punchagan:
Keeping the issue open until we have some kind of a test for this, based on @art-w's comment [here](https://github.com/ocurrent/current-bench/pull/299#issue-1128617362):

> see https://github.com/ocurrent/current-bench/issues/297 ... but I would like to keep this issue open until we have a way of checking that this bug will not reappear next time we touch the pipeline!:

PR #299, opened 2022-02-09
Fix github status update: wait for benchmark to terminate

art-w:
A fix for production where benchmarks get a green "passed" flag on github, even when they are still running.

(see https://github.com/ocurrent/current-bench/issues/297 ... but I would like to keep this issue open until we have a way of checking that this bug will not reappear next time we touch the pipeline!)

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -130,7 +130,8 @@ let pipeline ~ocluster ~conninfo ~repository env =
         Logs.err (fun log -> log "Error in %s stage: %s\n\n" stage m);
         Storage.record_stage_failure ~stage ~serial_id ~reason:m ~conninfo
     | _ -> ()
-  in
+  and+ () = ocluster_worker
+  and+ _ = output in
   ()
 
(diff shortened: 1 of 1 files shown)


gs0510:
Ahh this is failing because `Cmdliner.Term` has been deprecated in the latest version. The Changelog is here: https://github.com/dbuenzli/cmdliner/blob/master/CHANGES.md#new-cmd-module-and-deprecation-of-the-term-evaluation-interface 😬

pipeline/lib/pipeline.ml
```diff
@@ -130,7 +130,8 @@ let pipeline ~ocluster ~conninfo ~repository env =
         Logs.err (fun log -> log "Error in %s stage: %s\n\n" stage m);
         Storage.record_stage_failure ~stage ~serial_id ~reason:m ~conninfo
     | _ -> ()
-  in
+  and+ () = ocluster_worker
```
gs0510, line 133: This looks good to me! Just to confirm, we bind the cluster worker and output so that we only send the state to Github once we have the output right? :)

art-w: Yes exactly!

PR #303, opened 2022-02-11
Run DB migrations on pipeline start

punchagan:
We previously did this through a`make` target, but I think this is more robust, ensuring that the required DB migrations have been run for the current version of the app that we are going to start running.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,6 @@ stop-production:
 .PHONY: redeploy-production
 redeploy-production: \
 	build-production \
-	run-migrations \
 	stop-production \
 	start-production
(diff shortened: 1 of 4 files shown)


punchagan:
> Maybe we can also add the db password env variable to the validate script? this way we catch all possible errors? :)

We already use the `?required` option and use the same env var as the password in the `--conn-info` argument to the pipeline. I think it's good enough, and I'm not sure the validate script will add any thing more, here, unless I'm missing something?

gs0510:
No you are not, I missed the `required`! thanks!

PR #331, opened 2022-03-04
Show only workers with valid runs in workers dropdown

punchagan:
Show only workers which are relevant for a specific (repo_id, pull_number) in the workers dropdown.


art-w (approved):
LGTM thanks!

Issue #312, opened 2022-02-18
Time scheduling


art-w:
We currently run a benchmark on every new commit on the default branch and PRs.  But there is also a need for daily, weekly and monthly reports to study long term progress, since it allows to run more intensive benchmarks that would be too slow and not representative on individual changes (see also https://github.com/ocurrent/current-bench/issues/23)

- [ ] We could extend the `production.conf` to enable this alternative time scheduling for repositories that requires it
- [ ] Since the really intensive benchmarks can block a worker for nearly a day, we could run them on a temporary [server with Equinoxe](https://github.com/maiste/equinoxe) to keep current-bench operational for the other repositories

shakthimaan:
This will be useful for `Sandmark` nightly runs as well, especially for the Custom Variant Support feature request https://github.com/ocaml-bench/sandmark/issues/274.

PR #335, opened 2022-03-10
Pipeline: allow running benchmarks on a cron-like schedule

art-w:
This PR adds an option to run benchmarks on a different schedule than "as soon as a PR changes" https://github.com/ocurrent/current-bench/issues/312 . In the `environment/production.conf`, the optional `"schedule"` field specifies a time-based cron-like schedule:

```json
{
  "repositories": [
    {
      "name": "local/local",
      "worker": "autumn",
      "image": "ocaml/opam:debian-11-ocaml-4.14",
      "schedule": "04:00:00"
    },
```

Here the schedule would be everyday at 4 o'clock in the morning. The parsing / time-computations are done by [timere](https://github.com/daypack-dev/timere), but the parsing isn't documented (?) so here are some examples:
- `"00:00:00"` for everyday at midnight
- `"00:00:00 or 12:00:00"` for twice a day
- `"sunday and 12:00:00"` for every sunday at 12 o'clock
- `"1st and 23:00:00"` for every 1st of each month at 23 o'clock

The code is a bit more complicated than expected. By default, OCurrent provides primitives for caching a result for a duration of time ("this benchmark is valid for 1 day"), but this makes the initial timings dependent on when we start the pipeline. I wanted something a bit more human-friendly where we know precisely when the benchmarks will run.

@shakthimaan > A limitation of our database is that a benchmark run is associated with a unique git commit: If the repository is not updated in between two time schedules, then the benchmark will not be re-run (since the pipeline considers that no change happened). This will be an issue for repositories that have external sources of changes other than their git history (the irmin reports, probably sandmark nightly?) I'll open a new issue for this, as currently the time-scheduling is only useful for running `Slow` benchmarks less frequently.

### DIFF
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -35,6 +35,8 @@ depends: [
   "postgresql"
   "rresult"
   "omigrate"
+  "timere"
+  "timere-parse"
   "alcotest" {with-test}
   "alcotest-lwt" {with-test}
 ]
--- a/pipeline/lib/clock.ml
+++ b/pipeline/lib/clock.ml
@@ -1,0 +1,58 @@
+open Lwt.Infix
+
+type t = string Current.t
+
+let next clock_descr =
+  let timeline = Timere.inter [ Timere.after (Timedesc.now ()); clock_descr ] in
+  match Timere.resolve timeline with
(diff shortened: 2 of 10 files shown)


punchagan:
Changes merged via #336

PR #345, opened 2022-03-30
Correctly handle carriage returns when counting line numbers

punchagan:
Previously, the line number counting code didn't increment line numbers when
encoutering carriage returns.  This is in-line with the line numbers displayed
by Unix tools like `cat -n`, for instance.  But browsers treat `\r` characters
differently -- `\r`, `\r\n` and `\n`, all display a newline in browsers.


art-w:
It looks very good, thanks!

PR #347, opened 2022-04-06
frontend: Use same color for a sub-metric across different graphs

### LINKED ISSUE #344: Show the same sub-metric in the same color across different metric graphs
For instance, `typing`, `generate`, etc. across different projects for the[ `ocaml/ocaml`](https://autumn.ocamllabs.io/ocaml/ocaml?worker=grumpy&image=bench.Dockerfile) metrics should show up in the same color .

punchagan:
To make it easier to parse and understand graphs for metrics like ocaml/ocaml
which show the compiler parsing, generation, typing times across different
projects, we use the same color for each of these sub-metrics.

Closes #344

image

### DIFF
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -136,6 +136,31 @@ module Benchmark = {
       ->Belt.Map.String.valuesToArray
     }, [benchmarkDataByTestName])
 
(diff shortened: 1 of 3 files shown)


art-w:
Thanks a ton!

PR #348, opened 2022-04-07
Ensure comparison benchmark size metrics use correct units

### LINKED ISSUE #322: Adjust units on comparison commit values too
As caught by @art-w, [here](https://github.com/ocurrent/current-bench/pull/320#discussion_r811880582), call AdjustMetrics.adjust on comparison benchmark values too.

punchagan:
Previously, the size units were not adjusted on the comparision benchmark data.
This commit ensures that the same units are used for each metric in the
comparision benchmark data as the unit for those metrics in the benchmark data.

*Before*
image

*After*
image

Closes #322


art-w:
Very good catch!

PR #349, opened 2022-04-11
Hide closed PRs in the UI

### LINKED ISSUE #346: Hide closed PRs
The old closed/merged PRs are always visible in the sidebar, which makes it harder to survey all the active benchmarks. These PRs also tend get stuck in a "running" state: https://autumn.ocamllabs.io/ocaml/dune/pull/5270?worker=autumn&image=ocaml%2Fopam%3Adebian-11-ocaml-4.13

punchagan:
Closes #346

### DIFF
--- a/pipeline/lib/storage.ml
+++ b/pipeline/lib/storage.ml
@@ -113,3 +113,24 @@ let record_success ~serial_id (db : Postgresql.connection) =
 
 let record_success ~serial_id ~conninfo =
   Db_util.with_db ~conninfo (record_success ~serial_id)
+
+let mark_closed_pull_requests ~open_pulls (db : Postgresql.connection) =
+  Logs.debug (fun log -> log "Updating open and closed pulls...");
+  let open_pr_query =
+    String.concat " OR "
+    @@ List.map
+         (fun (repo_id, pull_number) ->
+           Fmt.str {|(repo_id = '%s' AND pull_number = %d)|} repo_id pull_number)
+         open_pulls
+  in
+  let query =
+    Fmt.str
+      {|UPDATE benchmark_metadata
+        SET is_open_pr = pull_number is NULL OR %s;
+      |}
+      open_pr_query
+  in
+  ignore (db#exec ~expect:[ Postgresql.Command_ok ] query)
+
+let mark_closed_pull_requests ~open_pulls ~conninfo =
+  Db_util.with_db ~conninfo (mark_closed_pull_requests ~open_pulls)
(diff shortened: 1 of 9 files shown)


pipeline/lib/storage.ml
```diff
@@ -113,3 +113,29 @@ let record_success ~serial_id (db : Postgresql.connection) =
...
+  let query =
+    Fmt.str
+      {|UPDATE benchmark_metadata
+        SET pull_closed = true
+        WHERE pull_number is NOT NULL;
+
+        UPDATE benchmark_metadata
+        SET pull_closed = false
+        WHERE %s;
+      |}
```
art-w, line 135: Perhaps this could be a one-shot update (rather than two), with something like:
```sql
UPDATE ...
SET pull_closed = ( pull_number is not null AND NOT ( equal_to_one_of_the_open_pr ) )
```
But this actually removes the `NULL` state for branches... I personally have a hard time with negations, what do you think of replacing `pull_closed` with an `is_open` boolean not null? (such that the main branch can be `is_open=true`, and keeping the intuition that "true is good" that works well with filters etc)

pipeline/db/migrations/20220411085604_add_pr_open_closed.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE benchmark_metadata
+ADD COLUMN pull_closed BOOLEAN DEFAULT FALSE;
```
art-w, line 2: I think life can be tiny bit easier in the long run if we can have a `BOOLEAN NOT NULL` here (for branches, `pull_closed=false` sounds fine as we would get the expected "don't hide them" result)

(`NULL` is a bit of an "I-don't-know" state, so there's always a risk of interpreting it wrongly in the code later + it requires extra code to handle trinary logic!)

art-w:
This looks great! :fire: :fire: :fire:
I left a few comments, but this is borderline subjective review!

art-w (approved):

punchagan:
Thanks for your thoughtful review, @art-w! I will address your comments tomorrow and update the PR.

pipeline/db/migrations/20220411085604_add_pr_open_closed.up.sql
punchagan: Yes, I agree having a `NOT NULL` condition here would simplify things!

pipeline/lib/storage.ml
punchagan: This makes sense too. Matching the flag with the data being fetched from the API would make things easier to understand. I'll change this!

punchagan:
image
image

art-w:
This looks great, thanks!

PR #351, opened 2022-04-15
Opam install test dependencies

art-w:
We've had a bit of a back-and-forth on this issue before... In general running the benchmarks requires more dependencies than the "default basic installation", so it's a better default if we also install the `{with-test}` libraries from the `opam` file (and not force the user to maintain a custom `*-bench.opam`)


punchagan:
This seems reasonable to me.

PR #352, opened 2022-04-15
Fix GitHub status flag

art-w:
The github status was :heavy_check_mark: even when the builds were failing... as the slack logs were interfering in the reporting of the error.

(also when testing the github integration locally, `ngrok` was complaining that the `-log` flag was renamed to `--log`)


punchagan:
Thanks for fixing this!

PR #356, opened 2022-04-22
Record commit messages and display them on hover on commit IDs

punchagan:
This commit depends on a modified version of [`ocurrent`](https://github.com/punchagan/ocurrent/tree/capture-commit-message) that fetches the commit messages from GitHub.  The PR temporarily points to that commit of `ocurrent`. I'm not sure if we should wait to merge this PR until we submit that change, or if there's a better way to go about this.

### DIFF
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -29,6 +29,7 @@ depends: [
   "ocluster-worker"  {= "dev"}
   "obuilder"         {= "dev"}
   "obuilder-spec"    {= "dev"}
+  "dockerfile"
   "duration"
   "fpath"
   "logs"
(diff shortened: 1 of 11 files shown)


art-w:
It looks very good (including the ocurrent patch!) :100: I think we can wait a bit to see if they are willing to include it officially :)

art-w (approved):

punchagan:
> It looks very good (including the ocurrent patch!) 100 I think we can wait a bit to see if they are willing to include it officially :)

Thanks @art-w for the review and your help with working on this. (I still haven't figured out the best workflow to make changes in `ocurrent` and use them with `current-bench`, but discussing some of this with you helped put some things in perspective.)

punchagan:
Updated this PR with a commit which pins the `ocurrent` version to the latest dev version. It also required adding an explicit dependency on `dockerfile` since `ocurrent` no longer depends on it.

art-w:
Thanks a lot!

PR #357, opened 2022-05-03
Install *-bench.opam in priority

art-w:
In order to run the irmin benchmarks on the raspberry pi 4, we need to skip the installation of `libirmin.opam` (which disables arm64)... so we must only install the `irmin-bench.opam` and its dependencies, not everything.


gs0510 (approved):

gs0510:
Looks good to me! :) thanks @art-w.

PR #358, opened 2022-05-10
WIP: Look for changed significantly metrics and notify authors

punchagan:
This PR looks for PRs where metrics have changed significantly and attempts to post messages to the corresponding Github PR to notify authors to check the metrics.

### DIFF
--- a/pipeline/lib/metric.ml
+++ b/pipeline/lib/metric.ml
@@ -1,0 +1,295 @@
+module Github = Current_github
+
+module MetricKey = struct
+  type t = string * string * string
+
+  let compare a b = Stdlib.compare a b
+end
+
+module BenchmarksData = Map.Make (MetricKey)
+
+let change_threshold = 0.01
+
+let parse_benchmark_data ~data =
+  let metrics =
+    Array.fold_left
+      (fun acc datum ->
+        match datum with
+        | _commit, benchmark_name, test_name, metrics ->
+            metrics
+            |> List.fold_left
+                 (fun acc
+                      ({ name; value; _ } : Current_bench_json.Latest.metric) ->
+                   let key = (benchmark_name, test_name, name) in
+                   BenchmarksData.add key value acc)
+                 acc)
+      BenchmarksData.empty data
+  in
+  let commit =
+    if Array.length data > 0
+    then
+      let commit, _, _, _ = data.(0) in
+      Some commit
+    else None
+  in
+  (metrics, commit)
+
+let sum_float_list xs = List.fold_left (fun acc x -> acc +. x) 0. xs
+
+let mean_float_list xs =
+  let n = List.length xs in
+  match n with 0 -> 0. | _ -> sum_float_list xs /. Float.of_int n
+
+let avg_of_value (v : Current_bench_json.Latest.value) =
+  match v with
+  | Float v -> v
+  | Floats vs -> mean_float_list vs
+  | Assoc vs -> (
+      vs
+      |> List.filter_map (function
+           | key, value when key = "avg" -> Some value
+           | _ -> None)
+      |> function
+      | [ avg ] -> avg
+      | _ -> 0.)
+
+let calc_diff v cv =
+  if cv <> 0.
+  then (v -. cv) /. cv *. 100.
+  else if v <> 0.
+  then (v -. cv) /. v *. 100.
+  else 0.
+
+type change = Change of float | New
+
+let find_changed_metrics ~metrics ~compare_metrics =
+  BenchmarksData.merge
+    (fun _ (value : Current_bench_json.Latest.value option)
+         (compare_value : Current_bench_json.Latest.value option) ->
+      match (value, compare_value) with
+      | Some value, Some compare_value ->
+          let delta =
+            calc_diff (avg_of_value value) (avg_of_value compare_value)
+          in
+          if Float.abs delta > change_threshold
+          then Some (Change delta)
+          else None
+      | Some _, None -> Some New
+      | _ -> None)
+    metrics compare_metrics
+
+module SeenBenchmarks = Set.Make (MetricKey)
+
+let format_changes ~changes =
+  let text_list, _ =
+    BenchmarksData.fold
+      (fun (bench_name, test_name, metric_name) change (acc, seen) ->
+        let bench_key = (bench_name, "", "") in
+        let test_key = (bench_name, test_name, "") in
+        let bench_seen = SeenBenchmarks.mem bench_key seen in
+        let test_seen = SeenBenchmarks.mem test_key seen in
+        let entry, seen =
+          match bench_seen with
+          | false ->
+              ( Fmt.str "\n\n## Benchmark: %s\n" bench_name,
+                SeenBenchmarks.add bench_key seen )
+          | true -> ("", seen)
+        in
+        let entry, seen =
+          match test_seen with
+          | false ->
+              ( Fmt.str "%s\n### Test: %s\n" entry test_name,
+                SeenBenchmarks.add test_key seen )
+          | true -> (entry, seen)
+        in
+        let change =
+          match change with
+          | New -> Fmt.str "- %s is a new metric" metric_name
+          | Change change -> Fmt.str "- %s changed by %.1f%%" metric_name change
+        in
+        let entry = Fmt.str "%s%s" entry change in
+        (entry :: acc, seen))
+      changes ([], SeenBenchmarks.empty)
+  in
+  String.concat "\n" (List.rev text_list)
+
+let commit_url ~repo_id ~commit =
+  let url = Fmt.str "https://github.com/%s/commit/%s" repo_id commit in
+  let short_sha = String.sub commit 0 7 in
+  Fmt.str "[%s](%s)" short_sha url
+
+let pull_url ~repo_id ~pull_number =
+  let url = Fmt.str "https://github.com/%s/pulls/%d" repo_id pull_number in
+  let pull = Fmt.str "#%d" pull_number in
+  Fmt.str "[%s](%s)" pull url
+
+let benchmarks_url ~repo_id ~worker ~docker_image =
+  let base_url = Repository.frontend_url () in
+  Fmt.str "%s/%s?worker=%s&image=%s" base_url repo_id worker docker_image
+
+let format_changes_message ~changes ~repository ~worker ~docker_image
+    ~main_commit =
+  let pull_number = Repository.pull_number repository |> Option.get in
+  let commit = Repository.commit_hash repository in
+  let repo_id = Repository.info repository in
+  let main_commit = Option.value ~default:"Unknown hash" main_commit in
+  let metrics_url = benchmarks_url ~repo_id ~worker ~docker_image in
+  let header =
+    Fmt.str
+      "%s (%s) changes the [metrics](%s) as follows in comparison to `main` \
+       (%s) when running on `%s (%s)`:"
+      (pull_url ~repo_id ~pull_number)
+      (commit_url ~repo_id ~commit)
+      metrics_url
+      (commit_url ~repo_id ~commit:main_commit)
+      worker docker_image
+  in
+  Fmt.str "%s\n%s" header (format_changes ~changes)
+
+let post_github_comment ~api ~repository ~pull_number ~text =
+  let open Lwt.Infix in
+  let repo_owner = Repository.owner repository in
+  let repo_name = Repository.name repository in
+  let variables =
+    [
+      ("owner", `String repo_owner);
+      ("name", `String repo_name);
+      ("pullNumber", `Int pull_number);
+    ]
+  in
+  let query =
+    {|
+       query ($owner: String!, $name: String!, $pullNumber: Int!) {
+         repository(owner: $owner, name: $name) {
+           pullRequest(number: $pullNumber) {
+             id
+           }
+         }
+       }
+     |}
+  in
+  let ( / ) a b = Yojson.Safe.Util.member b a in
+  let prId =
+    Github.Api.exec_graphql api query ~variables >|= fun json ->
+    let open Yojson.Safe.Util in
+    let repo = json / "data" / "repository" in
+    let id = repo / "pullRequest" / "id" in
+    id |> to_string
+  in
+  let _ =
+    prId >|= fun id ->
+    let variables = [ ("subjectId", `String id); ("body", `String text) ] in
+    let mutation =
+      {|
+         mutation ($body: String!, $subjectId: ID!) {
+           addComment(input:{body: $body, subjectId: $subjectId}) {
+             commentEdge {
+               node {
+                 body
+               }
+             }
+           }
+         }
+       |}
+    in
+    Github.Api.exec_graphql api mutation ~variables >|= fun json ->
+    let open Yojson.Safe.Util in
+    let body =
+      json / "data" / "addComment" / "commentEdge" / "node" / "body"
+      |> to_string
+    in
+    print_endline @@ body
+  in
+  ()
+
+module NotifyGithub = struct
+  type t = {
+    conninfo : string;
+    repository : Repository.t;
+    github_api : Github.Api.t option;
+    worker : string;
+    docker_image : string;
+    pull_number : int;
+  }
+
+  let id = "notify-github"
+  let pp h (key, _) = Fmt.pf h "notify-github %s" key
+
+  module Key = struct
+    type t = string
+
+    let digest t : string = t
+  end
+
+  module Value = Current.Unit
+  module Outcome = Current.String
+
+  let auto_cancel = true
+
+  let publish
+      { conninfo; repository; worker; docker_image; github_api; pull_number; _ }
+      job _worker_job_id () =
+    let open Lwt.Infix in
+    Current.Job.start job ~level:Current.Level.Above_average >>= fun () ->
+    Logs.info (fun log -> log "Looking for significantly changed metrics...");
+    match
+      Storage.get_main_branch_metrics ~conninfo ~repository ~worker
+        ~docker_image
+    with
+    | Some result_main, Some result -> (
+        let compare_metrics, main_commit =
+          parse_benchmark_data ~data:result_main
+        in
+        let metrics, _ = parse_benchmark_data ~data:result in
+        let changes = find_changed_metrics ~metrics ~compare_metrics in
+        match github_api with
+        | Some api when not (BenchmarksData.is_empty changes) ->
+            let text =
+              format_changes_message ~changes ~repository ~worker ~docker_image
+                ~main_commit
+            in
+            let () = post_github_comment ~api ~repository ~pull_number ~text in
+            Lwt.return (Ok text)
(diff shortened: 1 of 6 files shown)


pipeline/lib/metric.ml
```diff
@@ -0,0 +1,134 @@
...
+                   BenchmarksData.find_opt name compare_benchmark_metrics
+                 in
+                 let delta =
+                   match compare_value with
+                   | Some compare_value -> (
+                       match (value, compare_value) with
+                       | Float v, Float cv when cv != 0. -> calc_diff v cv
+                       | Floats vs, Floats cvs ->
+                           let v = sum_float_list vs in
+                           let cv = sum_float_list cvs in
```
art-w, line 58: Ah I think the current meaning of `Floats` is "repeated benchmarks" (so comparing the sum without dividing by the length might cause some issues if the user repeats the benchmark more in the future)

pipeline/lib/storage.ml
```diff
@@ -135,3 +135,33 @@ let mark_closed_pull_requests ~open_pulls (db : Postgresql.connection) =
...
+    (db : Postgresql.connection) =
+  let repo_id = Db_util.string @@ Repository.info repository
+  and commit = Db_util.string @@ Repository.commit_hash repository
+  and worker = Db_util.string @@ worker
+  and docker_image = Db_util.string @@ docker_image in
+  let main_commit =
+    Fmt.str
+      {|(SELECT commit FROM benchmarks
+       WHERE pull_number is NULL AND repo_id=%s AND worker=%s AND docker_image=%s
+       ORDER BY -EXTRACT(epoch FROM run_at) LIMIT 1)
```
art-w, line 149: Nitpick: would `ORDER BY run_at DESC` work? In any case well done, this query is great to handle the multiple entries associated with a single commit!

pipeline/lib/storage.ml
```diff
@@ -135,3 +135,33 @@ let mark_closed_pull_requests ~open_pulls (db : Postgresql.connection) =
...
+    Fmt.str
+      {|SELECT commit, benchmark_name, test_name, metrics FROM benchmarks
+       WHERE commit = %s AND repo_id=%s AND worker=%s AND docker_image=%s;
+       |}
+  in
+  let query = query_tmpl main_commit repo_id worker docker_image in
+  let result_main = db#exec query in
+  let query = query_tmpl commit repo_id worker docker_image in
+  let result = db#exec query in
+  (result_main#get_all, result#get_all)
```
art-w, line 163: Nitpick: it might be more verbose, but I think "parsing" the sql results at this stage would help the typechecker catch bugs in the long run... The issue with a `string array array` output is that if the query is updated and returns more/less columns, then the typechecker is not going to complain that we broke the users of this function. The "parsing" can be as simple as:
```ocaml
Array.map (function [| commit ; bench_name ; test_name ; metrics |] ->
                      (commit, bench_name, test_name, metrics) ...
```
(but we might also want to `metrics |> Yojson.Safe.from_string |> Current_bench_json.Latest.metrics_of_json []` to ensure a coherent typing)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,134 @@
...
+      match datum with
+      | [| _commit; benchmark_name; test_name; metrics |] ->
+          let key = Fmt.str "%s__%s" benchmark_name test_name in
+          let parsed_metrics = Yojson.Safe.from_string metrics in
+          BenchmarksData.add key parsed_metrics acc
+      | _ -> acc)
+    BenchmarksData.empty data
+
+let sum_float_list xs = List.fold_left (fun acc x -> acc +. x) 0. xs
+let calc_diff v cv = Float.abs (v -. cv) /. cv *. 100.
```
art-w, line 16: Perhaps this function should be responsible for checking `cv <> 0.0` ? (otherwise this responsibility is pushed to the caller, and if we ever change the logic of `calc_diff` such that this precondition is not required anymore... For example I think we might have to do `Float.abs cv > 0.0001` rather than `cv <> 0.0` because division by near zero can quickly yield huge numbers even though we don't really care about this tiny metric)

pipeline/lib/storage.ml
punchagan: Yes, that works too. Thanks!

pipeline/lib/metric.ml
punchagan: Ah, yes. I think using the mean is a safe way to do this. Thanks for catching this!

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,134 @@
...
+let find_avg vs =
+  vs
+  |> List.filter_map (function
+       | key, value when key = "avg" -> Some value
+       | _ -> None)
+  |> function
+  | [ avg ] -> avg
+  | _ -> 0.
+
+let change_threshold = 0.01
```
gs0510, line 27: Is this how much change we are okay with? :) Perhaps along with the other config, we can ask projects for the thresholds for each benchmark they are okay with (just a passing thought, i don't know how annoying it will be 😬 )

pipeline/lib/metric.ml
punchagan: Makes sense. Fixed this! :+1:

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,134 @@
+module Github = Current_github
+module BenchmarksData = Map.Make (String)
+
+let parse_benchmark_data ~data =
+  Array.fold_left
+    (fun acc datum ->
+      match datum with
+      | [| _commit; benchmark_name; test_name; metrics |] ->
+          let key = Fmt.str "%s__%s" benchmark_name test_name in
+          let parsed_metrics = Yojson.Safe.from_string metrics in
```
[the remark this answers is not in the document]
punchagan: The flattening works nicely for comparison, but we'd probably need to group the metrics by bench_name and test_name when generating the notification message. But, I think it makes the code a little easier to understand with the flattened benchmarks map.

pipeline/lib/storage.ml
```diff
@@ -135,3 +135,53 @@ let mark_closed_pull_requests ~open_pulls (db : Postgresql.connection) =
...
+  let query = query_tmpl commit repo_id worker docker_image in
+  let result = db#exec query in
+  match (result_main#get_all, result#get_all) with
+  | result_main, result ->
+      (Some (parse_result result_main), Some (parse_result result))
+  | exception exn ->
+      Logs.err (fun log ->
+          log "Could not fetch metrics for comparison %s:%s\n%a" repo_id commit
+            Fmt.exn exn);
+      (None, None)
```
art-w, line 183: I think that re-raising the exception would work, and allow ocurrent and slack to notify us that there's an issue :)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,100 @@
+module Github = Current_github
+
+module MetricKey = struct
+  type t = string * string * string
+
+  let compare a b = compare a b
```
art-w, line 6: This `MetricKey` is a very very good idea! :) (for this kind of comparison, I like to do `let compare = Stdlib.compare` to make it obvious + not run into issues if another `compare` enters the scope)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,100 @@
...
+                let v = mean_float_list vs in
+                let cv = mean_float_list cvs in
+                calc_diff v cv
+            | Assoc vs, Assoc cvs -> calc_diff (find_avg vs) (find_avg cvs)
+            | _ -> 0.)
+        | _ -> 0.
+      in
+      match delta with
+      | _ when Float.abs delta > change_threshold -> Some delta
+      | _ -> None)
```
art-w, line 62: To clarify some earlier comment, I think I would go with something like this:
```ocaml
BenchmarksData.merge
  (fun _ value compare_value -> match value, compare_value with
    | Some value, Some compare_value ->
      let delta = calc_diff (mean value) (mean compare_value) in
      if Float.abs delta > change_threshold then Some delta else None
    | _ -> None)
```
We might want to report the "missing benchmarks" (with `| Some _, None -> ...`, but no delta condition this time)

punchagan: This makes sense. 

I don't know if we should also flag the case where `cv = 0`; `calc_diff` currently returns `0.` which would drop the metric when comparing against the threshold.

art-w: Yeah, zero will always be a bit weird... Perhaps the formula could be "is `cv` within ±10% of `v`" ? (to avoid the division)

punchagan: Yeah, that seems like a reasonable way to handle this! :+1:

punchagan: I changed `calc_diff` to divide with whichever value is non-zero (`a` or `b`). It might be nice to display the %age change in the GitHub comment, so I took this approach.

punchagan:
A sample bot message/comment would look like this: 

------------

[#2](https://github.com/local/local/pulls/2) ([40a7f39](https://github.com/local/local/commit/40a7f39023ef3fda09e43771914cd441274c9ee3)) changes the metrics as follows in comparison to `main` ([b5a7c3b](https://github.com/local/local/commit/b5a7c3bd1ac457aa800e0952bc954a1f4558a87d)) when running on `autumn (ocaml/opam:debian-11-ocaml-4.13)`:


## Benchmark: default

### Test: bench_1_test_1
- grammarFunctor/generate changed by 42.857143%
- grammarFunctor/parsing changed by 5.263158%
- grammarFunctor/typing changed by -28.571429%
- mbs_per_sec changed by -4.522613%
- ops_per_sec changed by -1.449275%

### Test: bench_1_test_2
- mbs_per_sec changed by -41.176471%
- ops_per_sec changed by 2.061856%
- time changed by -9.057971%
- yo is a new metric

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,230 @@
...
+          match test_seen with
+          | false ->
+              ( Fmt.str "%s\n### Test: %s\n" entry test_name,
+                SeenBenchmarks.add test_key seen )
+          | true -> (entry, seen)
+        in
+        let change =
+          if Float.is_nan change
+          then Fmt.str "- %s is a new metric" metric_name
+          else Fmt.str "- %s changed by %f%%" metric_name change
```
art-w, line 104: What do you think of `%.0f` or `%.1f` ? (not sure that we need super precise percentages)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,230 @@
...
+               commentEdge {
+                 node {
+                   body
+                 }
+               }
+             }
+           }
+         |}
+    in
+    Github.Api.exec_graphql api mutation ~variables >|= fun json ->
```
art-w, line 185: One thing I'm afraid of with OCurrent is that side effects can trigger more than once (notably after a restart of the pipeline). This graphql mutation may or may not be concerned (because we already do some stuff to avoid this...) but in doubt, could you look at [`Current_cache`](https://ocurrent.github.io/ocurrent/current/Current_cache/index.html) in order to allow only one comment per PR? :) (... I really really don't want us to spam)

art-w:
So cool! Would it be possible to add a setting per repository to opt-in this feature? (I would like to test it in production on safe repositories before enabling it for everyone ^^' )

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,230 @@
...
+  let repo_id = Repository.info repository in
+  let main_commit = Option.value ~default:"Unknown hash" main_commit in
+  let header =
+    Fmt.str
+      "%s (%s) changes the metrics as follows in comparison to `main` (%s) \
+       when running on `%s (%s)`:"
+      (pull_url ~repo_id ~pull_number)
+      (commit_url ~repo_id ~commit)
+      (commit_url ~repo_id ~commit:main_commit)
+      worker docker_image
```
art-w, line 135: Can we add a quick link to the benchmarks webpage for this worker/image? (otherwise it's a bit painful to access it ^^')

pipeline/lib/metric.ml
punchagan: I've tried to implement caching with `Current_cache` based on the docs, but I think it could use a careful look @art-w. I really appreciate your intent of not spamming and I'm fully onboard. :)

punchagan:
image

I see some new empty boxes added in the graph, and with an initial look at the API, I wasn't sure how to hide or label them. Could you help with that, @art-w ? :see_no_evil:

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,313 @@
...
+            Some repo_config
+        | _ -> None)
+      repo_configs
+  in
+  let pull_number = Repository.pull_number repository in
+  match (pull_number, should_notify) with
+  | Some pull_number, Some _ ->
+      let job_id = Current.return (Some job_id) in
+      let _ =
+        notify ~conninfo ~repository ~github_app ~worker ~docker_image
```
art-w, line 309: Given my understanding of ocurrent, I think the `let _ =` works, but is a bit sneaky! (a `Current.t` node is attached to the pipeline, yet the type of `notify_metric_changes` doesn't reflect it :/ ). Could we not ignore the outcome? (so that we would be able to report any errors to slack in this stage?)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,313 @@
...
+              let+ installations = Github.App.installations app in
+              let apis = List.map Github.Installation.api installations in
+              let api = match apis with [] -> None | api :: _ -> Some api in
+              match api with
+              | Some api ->
+                  post_github_comment ~api ~repository ~pull_number ~text
+              | None -> ()
+            in
+            Lwt.return (Ok text)
+        | _ -> Lwt.return (Ok "No Github app configured or no changes found"))
```
art-w, line 262: We can probably afford to differentiate between a bad configuration and the lack of changes to ease debugging in the future :P

```ocaml
| Some _ -> Ok "no changes"
| _ -> Error "no github app"
```

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,313 @@
...
+    pull_number : int;
+  }
+
+  let id = "notify-github"
+  let pp h (key, _) = Fmt.pf h "notify-github %s" key
+
+  module Key = struct
+    type t = string
+
+    let digest t : string = t
```
art-w, line 221: From memories, OCurrent uses the key to determine if an output has already been done: I'm still very afraid of spamming PRs, and I think the code currently sends one message per execution (job_id). Could we try it with one comment max per PR? (so that it acts as a warning, but doesn't keep on re-sending similar messages on further commits)

pipeline/lib/metric.ml
art-w: (... but if this is changed, there's some extra-difficulties in order to not (?) send an enormous message when the execution crashed and produced no json!)

pipeline/lib/metric.ml
punchagan: Great point! When I looked at your comment, I had a vague feeling that I did think of this, but couldn't find the code for it immediately with the limited context GitHub shows for the comments. (It has been a while since I worked on this :grin: )

I looked at the code again to find that I have [this "hack"](https://github.com/punchagan/current-bench/blob/bbc404de91116ca2f69d3ce7ae1d6825bf3bec7b/pipeline/lib/metric.ml#L292) to make the `job_id` to be associated with a commit. This means we'd only send a link once per commit, but we could send multiple notifications if commits are pushed onto a PR over time. I've seen other tools like code coverage tools work similarly, and I think this shouldn't feel too spammy to our users?

pipeline/lib/metric.ml
punchagan: I'm not sure what the best way out here is. I pushed a commit to use an `ignore` instead of `let _`, but not sure if that's what we want here.

pipeline/lib/metric.ml
art-w: Oh right, I missed that! (and thought it was the ocurrent `job_id`) ... Alright let's try it like that then!

pipeline/lib/metric.ml
art-w: Ha but it's very much the same as before with `ignore` :P 
If `job_id` was defined as dependent on the worker output, then it wouldn't be available until the worker terminates (something like `let job_id = Current.map (fun _ -> job_id) worker_output`) ... which would allow to `and+ () = Metric.notify...` rather than `ignore` :)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,313 @@
...
+          when worker = repo_config.worker && docker_image = repo_config.image
+          ->
+            Some repo_config
+        | _ -> None)
+      repo_configs
+  in
+  let pull_number = Repository.pull_number repository in
+  match (pull_number, should_notify) with
+  | Some pull_number, Some _ ->
+      let job_id = Current.return (Some job_id) in
```
art-w, line 309: Why is `job_id` an option? ^^

pipeline/lib/metric.ml
punchagan: Ah, yes. That makes sense! Let me try that.

pipeline/lib/metric.ml
punchagan: I've now changed the "job_id" slightly to include the worker and the docker_image, so that we are able to send messages for different worker+docker_image combinations for the same commit_hash.

pipeline/lib/metric.ml
punchagan: Yep, that's not required. Removed.

pipeline/lib/metric.ml
punchagan: I think my current change should be close to what you suggest. Could you take a look again? :)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,305 @@
...
+        let changes = find_changed_metrics ~metrics ~compare_metrics in
+        match github_app with
+        | Some app when not (BenchmarksData.is_empty changes) ->
+            let text =
+              format_changes_message ~changes ~repository ~worker ~docker_image
+                ~main_commit
+            in
+            let open Current.Syntax in
+            let _ =
+              let+ installations = Github.App.installations app in
```
art-w, line 253: Ha I nearly missed that one :D ! Can't we pass it through as part of the type `t`?

punchagan: Yes! I'm glad you caught this; this piece of code looks much cleaner now. :)

pipeline/lib/metric.ml
```diff
@@ -0,0 +1,311 @@
...
+let notify ~conninfo ~repository ~github_app ~worker ~docker_image ~pull_number
+    job_id =
+  match github_app with
+  | Some app ->
+      let open Current.Syntax in
+      Current.component "notify-github"
+      |> let> installations = Github.App.installations app and> job_id in
+         let github_api =
+           List.find_map
+             (fun install -> Some (Github.Installation.api install))
```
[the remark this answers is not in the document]
punchagan: That's a great idea. It made the whole code much simpler! Thanks. :)

art-w:
Thanks a lot for all the changes!

PR #359, opened 2022-05-16
calculate execution time of functions

Zineb-Ada:
The purpose of this PR is to allow users to calculate the execution time of their functions accurately and noiselessly. We have implemented two different functions to calculate this execution time, one with the bechamel library and the other with Current time (Sys.time).
The interest is to facilitate the use of these functions. Indeed, the user only has to replace in “cobench-examples/example.ml” :
  - “test_fact()” and “test_fibo()” by its functions
  - The quota which is an optional parameter. The quota parameter is the maximum of time allowed. the user can adjust it by increasing or decreasing the time allocated by bechamel or he can also remove it because it is 1 second by default.
  - The strings by the names he wants


cobench/cobench.ml
```diff
@@ -83,3 +84,55 @@ module Remote = struct
...
+  ()
+
+let time_loop ?(quota = 1.) name1 name2 t =
+  let t_time = Sys.time () in
+  while Sys.time () -. t_time < quota do
+    let timing = time t in
+    pr_bench name1 name2 timing
+  done
+
+let bench ?(quota = 1.) name1 name2 t =
```
punchagan, line 133: Would it be better to call these names `test_name` and `metric_name`? And the function argument to be called `f`? `t` is used as the default type name for a module in `OCaml` and since we are also in the context of timing and benchmarking `t` seems like a confusing name.

cobench/cobench.ml
```diff
@@ -83,3 +84,55 @@ module Remote = struct
...
+  let timing = time t in
+  if timing > 0.1 then (
+    pr_bench name1 name2 timing;
+    time_loop ~quota name1 name2 t)
```
punchagan, line 137: We already timed the function once to compute the value of `timing` to decide whether we want to use `bechamel` or `Sys.time`. We are running the `time_loop` here irrespective of how long the first run took. I think it makes sense to run the `time_loop` only if we haven't already exceeded the `quota`? If not, for long running functions, we could spending a lot more time than the `quota`.

punchagan (commented):
Thank you, @Zineb-Ada, for working on this! This would be a really helpful feature for the `cobench` library. The PR looks good in general, except for some minor changes that I thought would improve this PR.

Zineb-Ada:
Thanks you for your return. I fixed it according to your recommendations.

cobench/cobench.ml
```diff
@@ -83,3 +84,55 @@ module Remote = struct
...
+      | None -> ()
+      | Some ts ->
+          let t = List.hd ts in
+          pr_bench test_name metric_name (t /. 1000000.))
+    timings;
+  ()
+
+let time_loop ?(quota = 1.) initial_timing test_name metric_name f =
+  let t_time = Sys.time () in
+  while Sys.time () -. t_time +. initial_timing < quota do
```
art-w, line 128: Nice! I would prefer if the function `time_loop` didn't take the new argument `initial_timing` (because we can do without it ^^). This while condition is the same as:

```ocaml
while Sys.time () -. t_time < quota -. initial_timing do
```

So perhaps the remaining `quota` could be adjusted at some point to take into account the `initial_timing` :)

cobench-examples/dune
```diff
@@ -0,0 +1,7 @@
+(executable
+ (name example)
+ (modules example)
+ (libraries cobench))
+
+(rule
+ (copy ../pipeline/lib/current_bench_json.ml current_bench_json.ml))
```
art-w, line 7: I think you can remove that `rule`, it's only useful within the `cobench` library (but shouldn't be required by the example)

cobench-examples/example.ml
```diff
@@ -0,0 +1,19 @@
...
+  !i
+
+let test_fibo () =
+  let rec fibonacci n =
+    match n with 0 -> 0 | 1 -> 1 | n -> fibonacci (n - 1) + fibonacci (n - 2)
+  in
+  fibonacci 10
+
+let bech_of_sys_fact = bench ~quota:1.1 "bench" "factorial" test_fact
+let bech_of_sys_fibo = bench ~quota:1.1 "bench" "fibonacci" test_fibo
```
art-w, line 19: Can you replace the names `bech_of_sys_...` by `let () = bench ...` ? (to emphasize that the `bench` function doesn't return anything useful)

cobench.opam
```diff
@@ -12,6 +12,7 @@ depends: [
   "yojson"
   "cohttp-lwt-unix"
   "odoc" {with-doc}
+  "bechamel" {with-test}
```
art-w, line 15: Can you remove the `{with-test}` here? (since bechamel is actually essential to the library, not just for testing it)

art-w:
Thanks a lot, this looks great!

PR #364, opened 2022-06-14
pipeline: Don't filter out local repository as stale repo

punchagan:
This commit fixes the bug introduced in
2ff8bb142205b15fcf8e43262515bd70cca5e303 which incorrectly filters out
"local/local" repository when filtering stale repositories.


art-w:
Merci ! ^^

PR #365, opened 2022-06-15
Improve json identification with FSA

### LINKED ISSUE #363: Improve json identification
We use a pseudo-parser to identify the json(s) inside the execution logs, then feed only those substrings to yojson for further analysis. However our custom parser is a bit too simple to work for everybody: It only checks that the "json" has correctly balanced parentheses `{}[]""` (= it's a bit too generous). As a result, we silently ignore the wrongly identified json when yojson crashes and are unable to confidently report the mistakes to the user (because it's most likely our fault).

- The ocaml compiler and ocamlformat can output lines that starts with a `{` but what follows is not json (for example https://github.com/ocurrent/current-bench/issues/321 )... A `{` should be followed by `"` or `}`, and our parser shouldn't recognize `{{{ this is not json }}}` and variations on this theme
- Since our pseudo-parser is so fragile, we restricted it to identify json only when it starts at the very beginning of the line (so right after a `\n`). It would be nice if we could be more flexible here! (because letting the user prefix its json output can enable him to `grep '^current-bench'` and so forth)

ElectreAAS:
This mainly resolves #363
I only added a small test to pinpoint the problem, I'll work on putting more tests in there in this PR

### DIFF
--- a/pipeline/tests/json_stream_test.ml
+++ b/pipeline/tests/json_stream_test.ml
@@ -49,7 +49,7 @@ let parse_two =
   ()
 
 let parse_wrong =
-  Alcotest_lwt.test_case_sync "parse wrong" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "parse json: invalid" `Quick @@ fun () ->
   let str = "{{{ this isn't json }}}" in
   let parsed = Json_stream.json_full str in
   let expect = [] in
@@ -57,7 +57,7 @@ let parse_wrong =
   ()
 
 let parse_wrong_longer =
-  Alcotest_lwt.test_case_sync "parse_wrong_longer" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "parse json: longer invalid" `Quick @@ fun () ->
   let str =
     {|{
       s =
@@ -74,7 +74,7 @@ let parse_wrong_longer =
   ()
 
 let parse_real_log =
-  Alcotest_lwt.test_case_sync "parse_real_log" `Quick @@ fun () ->
+  Alcotest_lwt.test_case_sync "parse json: real log" `Quick @@ fun () ->
   let str =
     {|> Start benchmarks on [fn¹].
     [                                        ] 0%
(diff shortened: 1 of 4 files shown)


pipeline/tests/json_stream_test.ml
```diff
@@ -48,4 +48,81 @@ let parse_two =
...
+  let parsed = Json_stream.json_full str in
+  let expect = [] in
+  Alcotest.(check (list parsed_location)) "jsons" expect parsed;
+  ()
+
+let parse_wrong_longer =
+  Alcotest_lwt.test_case_sync "parse_wrong_longer" `Quick @@ fun () ->
+  let str =
+    "{\n\
+    \      s =\n\
```
art-w, line 63: I don't think ocaml supports `\` at the beginning of a line (?) (but it does work at the end to skip the line break)

art-w:
Very cool! Thanks for adding more tests, the automata is quite involved :) Can we also have an example of "wrong json followed by valid json" to check that it recovers? (also json anywhere in a line since you also fixed that!)

pipeline/tests/json_stream_test.ml
ElectreAAS: I didn't add those, ocamlformat did. Interestingly, it doesn't do that when I use the full-escape form of strings `{| foo |}`

art-w:
Thanks!

punchagan:
@ElectreAAS I tried testing this with the `local-test-repo/Makefile` and it seems to error out.  Could you take a look? 

You could trigger a build using this Makefile by running `make local-make-bench`, for instance.

ElectreAAS:
@punchagan I don't want to be that person, but... it works on my machine :laughing:
if I `cd local-test-repo; make bench` as advised by @art-w, it seems to work...? It prints a lot of things, but looking at the Makefile it doesn't do anything, just calls to `echo`. What errors do you have?

If at the root I `make local-make-bench` it will try to git commit stuff, but will stop since there are no changes... Very weird makefiles

ElectreAAS:
By the way the only thing `make local-make-bench` does is this: `cd local-test-repo/; git add .; git commit --amend -m "New commit: $$(date)"; cd ..`, which is... not very useful?

punchagan:
Thanks for adding more context to my comment, @art-w !

PR #366, opened 2022-06-20
Pin prometheus-app=1.1 in cb-worker.opam and cluster/Dockerfile

punchagan:


art-w:
Ha @ElectreAAs had something similar when fighting the CI, but it must have been lost at some point. Thanks!

PR #368, opened 2022-06-21
Fix FSA parsing numbers incorrectly in json

### LINKED ISSUE #367: JSON parser fails when parsing numbers in long objects
@ElectreAAS I tried testing this with the `local-test-repo/Makefile` and it seems to error out.  Could you take a look? 

You could trigger a build using this Makefile by running `make local-make-bench`, for instance.

_Originally posted by @punchagan in https://github.com/ocurrent/current-bench/issues/365#issuecomment-1160223511_

ElectreAAS:
Resolves #367

### DIFF
--- a/pipeline/lib/json_stream.ml
+++ b/pipeline/lib/json_stream.ml
@@ -122,7 +122,7 @@ let json_step_aux stack chr =
     | InNumber :: InArray :: st, ',' -> BeforeValue :: InArray :: st
     | [ InNumber ], '}' -> raise Finished_JSON
     | InNumber :: st, '}' -> st
-    | InNumber :: st, ',' -> AfterValue :: st
+    | InNumber :: st, ',' -> BeforeID :: st
     | InNumber :: _, _ -> raise Invalid_JSON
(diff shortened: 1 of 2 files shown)


art-w:
To avoid another surprise in the future, could we add another unit test that runs `make bench` in `local-test-repo` and check that the parser succeeds on the output? :) (it could be a more complex test that checks that the identified substrings parse correctly with `Yojson`, and are accepted by `Current_bench_json`)

ElectreAAS:
Do we have a way to call `make` from inside ocaml? Otherwise the only thing that I can do is copy the json from the Makefile to a unit test

art-w:
Yes there is [`Unix.open_process_in`](https://v2.ocaml.org/api/Unix.html#VALopen_process_in) to run arbitrary shell commands (and grab their output... and perhaps [`Unix.chdir`](https://v2.ocaml.org/api/Unix.html#VALchdir) would also be useful here). I kind of like the idea that the unit tests are kept in sync with the `local-test-repo` rather than a static file, such that we don't forget to update them :)

art-w:
Thanks!

PR #371, opened 2022-07-04
Only run benchmarks for tagged PRs

art-w:
This adds a new optional setting for workers (following the request: https://github.com/ocaml/merlin/pull/1485#pullrequestreview-1018443932 ) in our `environments/*.conf`:

```json
{
  "repositories": [
    {
      "name": "owner/repo",
      "if_label": "bench"
    },
```

When `if_label` is configured, the benchmarks will run only for PRs that have this flag/label/tag enabled on github. Benchmarks still run as normal for the main/master branch (otherwise PR would have zero context to decide if a regression happened!) or when no `if_label` has been set (which is currently the case for every projects).

Since the label is configurable per worker, it allows to toggle only some workers off (for example "rpi-bench" could be an optional tag, as the raspberry benchmarks are much slower than the normal workers).

PS: I had to upgrade ocamlformat to the latest 0.22.4, sorry for the noise!


ElectreAAS:
I force-pushed a rebase of the branch on `main`. It doesn't compile anymore, but at least it's not because of a bug that we already fixed on main.

ElectreAAS:
> I force-pushed a rebase of the branch on `main`. It doesn't compile anymore, but at least it's not because of a bug that we already fixed on main.

My rebase was very bad so I just undid it. Will rebase properly soon.

ElectreAAS:
Thanks to ocurrent merging [my PR](https://github.com/ocurrent/ocurrent/pull/402) this one is ready for a full review. I tested it locally and everything seems to be in order.

art-w:
Thanks a lot!

PR #373, opened 2022-07-08
Removed persistance of 'failed' state after successful rebuild

### LINKED ISSUE #372: Failed state persists after successful rebuild
Sometimes the benchmarks fails out of bad luck (e.g. network failures during the docker build). One can hit the "Rebuild" button on the build logs page which will fix the misstep. However, the current-bench database still remembers the initial failure with the `failed=true` and `reason` column -- and the frontend keeps displaying the commit as failing.

(It's probably not ideal that the database has both `success`, `failed` and `cancelled` columns, I'm not sure how to interpret them when more than one is `true`!)

ElectreAAS:
Resolves #372.
In the future, using an enum instead of separate booleans seems more appropriate.


pipeline/lib/storage.ml
```diff
@@ -108,7 +112,10 @@ let record_success ~serial_id (db : Postgresql.connection) =
   let query =
     Fmt.str
       {|UPDATE benchmark_metadata
-        SET success = true
+        SET success = true,
+            failed = false,
+            cancelled = false,
+            reason = ''
```
art-w, line 118: I don't think it really matters, but I guess `reason` was supposed to be `NULL` when unset

ElectreAAS:
@punchagan I had to upgrade a few packages and pins, Arthur told me you'd find this interesting

art-w:
Thanks!

PR #380, opened 2022-08-23
Update obuilder dependency

art-w:
The production server rebooted during the break (?) and was unable to restart the services automatically following some api changes in our dependencies... I'm tempted to switch to git submodules as in the other CI projects, WDYT?
For similar reasons, the custom dockerfiles were now produced with the wrong syntax `COPY ["--chown=opam:opam ./*.opam", "./"]` instead of the correct `COPY --chown=opam:opam ["./*.opam", "./"]`


cluster/Dockerfile
```diff
@@ -2,8 +2,8 @@ FROM ocaml/opam:debian-ocaml-4.13 AS build
 RUN sudo apt-get update && sudo apt-get install libc6-dev libev-dev capnproto m4 pkg-config libsqlite3-dev libgmp-dev libffi-dev -y --no-install-recommends
 RUN opam remote add origin 'https://opam.ocaml.org' --all-switches && opam remote remove default && opam update
 RUN opam install dune>=2.8 ppx_expect>=v0.14.1 prometheus ppx_sexp_conv dune-build-info lwt>=5.4.1 capnp-rpc-lwt>=1.2 capnp-rpc-net capnp-rpc-unix>=1.2 extunix>=0.3.2 logs conf-libev digestif>=0.8 fpath lwt-dllist prometheus-app=1.1 cohttp-lwt-unix sqlite3 psq mirage-crypto>=0.8.5 ppx_deriving ppx_deriving_yojson astring fmt>=0.8.9 cmdliner tar-unix>=2.0.0 yojson sexplib sha
-RUN mkdir src && cd src && git clone 'https://github.com/ocurrent/ocluster' && opam pin -yn ocluster
-RUN cd src && git clone 'https://github.com/ocurrent/obuilder' && opam pin -yn obuilder
```
tmcgilchrist, line 6: @art-w could this use ocurrent's version of ocluster and obuilder?

art-w: Yes I don't think the custom pin is required there anymore! I'll close this as https://github.com/ocurrent/current-bench/pull/389/ is more uptodate :)

PR #387, opened 2022-10-19
Added prod.sh script and streamlined dev.sh usage in the Makefile

ElectreAAS:
This PR is the second one in the series of committing what was on the prod server.
The script `prod.sh` was added, and I renamed `dc.sh` to `dev.sh` to make it clear that one uses the development environment, while the other uses the production env. I also used those scripts in the Makefile wherever needed, to simplify future maintenance work.


ElectreAAS:
For more insight as to what the scripts do, check out #168

punchagan:
Thanks for making these changes, @ElectreAAS !

PR #389, opened 2022-10-20
Fix pins of ocluster & ocurrent in Dockerfile & .opam file

ElectreAAS:
This PR is the third one in the series of committing/cleaning what was on the prod server.
~~There were pins on art-w/ocluster even though that repo isn't actively maintained.~~
Part of the pin actually made sense, as the package `ocluster-worker` isn't merged into the main branch yet, so I added a comment to specify when to remove the pin.

Another weird thing is the line `sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam`. This (AFAIK) forces the used version of opam to be 2.1 instead of the default 2.0. Why this change was made in the first place I don't really know, but there might be someone down the line assuming opam version to be 2.1 so better not change this without full knowledge

### DIFF
--- a/cluster/Dockerfile
+++ b/cluster/Dockerfile
@@ -1,9 +1,9 @@
 FROM ocaml/opam:debian-ocaml-4.14 AS build
 RUN sudo apt-get update && sudo apt-get install libc6-dev libev-dev capnproto m4 pkg-config libsqlite3-dev libgmp-dev libffi-dev -y --no-install-recommends
 RUN opam remote add origin 'https://opam.ocaml.org' --all-switches && opam remote remove default && opam update
+RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
 RUN opam install dune>=2.8 ppx_expect>=v0.14.1 prometheus ppx_sexp_conv dune-build-info lwt>=5.4.1 capnp-rpc-lwt>=1.2 capnp-rpc-net capnp-rpc-unix>=1.2 extunix>=0.3.2 logs conf-libev digestif>=0.8 fpath lwt-dllist prometheus-app=1.1 cohttp-lwt-unix sqlite3 psq mirage-crypto>=0.8.5 ppx_deriving ppx_deriving_yojson astring fmt>=0.8.9 cmdliner tar-unix>=2.0.0 yojson sexplib sha
-RUN mkdir src && cd src && git clone 'https://github.com/ocurrent/ocluster' && opam pin -yn ocluster
-RUN cd src && git clone 'https://github.com/ocurrent/obuilder' && opam pin -yn obuilder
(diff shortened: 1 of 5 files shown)


punchagan:
@art-w Could you look at this PR since the changes are very similar to what you did in #380, and I don't understand what the best way forward here is.

cluster/Dockerfile
```diff
@@ -1,9 +1,9 @@
 FROM ocaml/opam:debian-ocaml-4.13 AS build
 RUN sudo apt-get update && sudo apt-get install libc6-dev libev-dev capnproto m4 pkg-config libsqlite3-dev libgmp-dev libffi-dev -y --no-install-recommends
 RUN opam remote add origin 'https://opam.ocaml.org' --all-switches && opam remote remove default && opam update
+RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
 RUN opam install dune>=2.8 ppx_expect>=v0.14.1 prometheus ppx_sexp_conv dune-build-info lwt>=5.4.1 capnp-rpc-lwt>=1.2 capnp-rpc-net capnp-rpc-unix>=1.2 extunix>=0.3.2 logs conf-libev digestif>=0.8 fpath lwt-dllist prometheus-app=1.1 cohttp-lwt-unix sqlite3 psq mirage-crypto>=0.8.5 ppx_deriving ppx_deriving_yojson astring fmt>=0.8.9 cmdliner tar-unix>=2.0.0 yojson sexplib sha
-RUN mkdir src && cd src && git clone 'https://github.com/ocurrent/ocluster' && opam pin -yn ocluster
-RUN cd src && git clone 'https://github.com/ocurrent/obuilder' && opam pin -yn obuilder
+RUN mkdir src && cd src && git clone 'https://github.com/ocurrent/ocluster' && git clone 'https://github.com/ocurrent/obuilder' && opam pin -yn ocluster obuilder
```
art-w, line 6: I'm running into some issues between obuilder and ocluster... It seems to work better with:
```
RUN mkdir src && cd src && git clone --recurse-submodules 'https://github.com/ocurrent/ocluster' && opam pin -yn ocluster
```

art-w:
Yes it finally works! Good job :) Do you know if the `omigrate` "dirty" column schema update is still required ? (for example would @punchagan run into the issue of the missing column?)

ElectreAAS:
> Do you know if the omigrate "dirty" column schema update is still required ?

Probably yes. I don't really know where to start to fix this in the database description side.
Just in case, here's the SQL line to be run if one where to encounter the "dirty" error locally:
`ALTER TABLE schema_migrations ADD dirty boolean NOT NULL DEFAULT false;`

art-w:
Thanks! Let's hope `omigrate` fixes their own schema migration, but otherwise it's not a big deal :)

PR #390, opened 2022-10-20
Simpler usage of dockerfile's API

ElectreAAS:
This PR is the fourth one (and hopefully one of the last ones) in the series of committing/cleaning what was on the prod server.
The `chown` optional argument to `copy` isn't documented at all in the latest version of the opam package `dockerfile`, but it is on the official docker's manual : https://docs.docker.com/engine/reference/builder/#copy


punchagan:
See related: #380

Issue #385, opened 2022-10-18
Provide JSON validation mechanism to current-bench consumers


shakthimaan:
As we benchmark more OCaml projects, we will need to ensure that the integration with current-bench is smooth. It will be good to provide JSON parsing capabilities and test to interested OCaml projects so that they can verify that their `make bench` target validates with current-bench.

ElectreAAS:
Solved with #422! :tada:

PR #393, opened 2022-11-07
Cleaner json handling

### LINKED ISSUE #265: Improve Json validation errors
The json parsing can raise hard to decipher errors when it doesn't find what it expects. It would be nice to have the json path to make the error easier to debug.

To give an example, the message `"Json: not a string"` gives no information of location: we could add `.results[1].name` to the message to indicate where the type error happened!

ElectreAAS:
Note: this PR is the same as #391, I just renamed the original branch.
This PR is the fifth and last one in the series of committing/cleaning what was on the prod server.
The change to exception handling in json parsing should make it way easier to debug, and eventually resolve #265 & #385.

### DIFF
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -283,7 +283,13 @@ let of_json json = Latest.of_json json
 let to_json t = Latest.to_json t
 
 let of_list jsons =
-  List.fold_left (fun acc json -> Latest.merge acc [ of_json json ]) [] jsons
+  List.fold_left
+    (fun acc json ->
+       match of_json json with
+       | t -> Latest.merge acc [t]
+       | exception _ -> acc)
+    []
+    jsons
 
 let to_list ts =
   List.map
--- a/pipeline/lib/current_bench_json.ml
+++ b/pipeline/lib/current_bench_json.ml
@@ -257,7 +257,7 @@ module V2 = struct
   let of_json t =
     let lines =
       match Json.get "lines" t with
-      | `Tuple [ `Int start; `Int end_ ] -> [ (start, end_) ]
+      | `Tuple [ `Int start; `Int finish ] -> [ (start, finish) ]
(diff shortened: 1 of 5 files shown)


pipeline/lib/current_bench_json.ml
```diff
@@ -278,16 +294,17 @@ end
...
-let to_json t = Latest.to_json t
-
 let of_list jsons =
-  List.fold_left (fun acc json -> Latest.merge acc [ of_json json ]) [] jsons
+  List.fold_left
+    (fun acc json ->
+      match Latest.of_json json with
+      | t -> Latest.merge acc [ t ]
+      | exception _ -> acc)
+    [] jsons
```
art-w, line 303: Perhaps we could do a `partition_map` like https://github.com/ocurrent/current-bench/blob/main/pipeline/lib/json_stream.ml#L263 to save the errors? :)
For context, the issue is that some logs have both invalid jsons + valid jsons, but the "invalid" jsons are just not intended for current-bench and so it's fine to ignore them:

https://github.com/ocurrent/current-bench/blob/main/pipeline/lib/current_bench_json.ml#L266

I haven't thought too much about it, but the absence of a toplevel `"results"` field could be used as a hint that the json wasn't for us? :)

ElectreAAS: > save the errors
Something like this?
```ocaml
List.fold_left
 (fun (jsons, exns) (json, range) ->
   try
     ( json
       |> Yojson.Safe.from_string
       |> json_merge_lines range
       |> Current_bench_json.Latest.of_json
       |> Current_bench_json.Latest.add jsons,
       exns )
   with exn ->
     (jsons, (json, range, Printexc.to_string exn) :: exns))
 ([], [])
```

art-w:
Thanks a lot! Can you deploy it in production to get rid of the irmin issues? :)

ElectreAAS:
Done!

PR #394, opened 2022-11-10
Added 2-hour timeout to benchmarks

### LINKED ISSUE #392: Add a timeout to benchmarks
Some benchmarks might take a long time to run and that's expected, but if one were to create a benchmark that is just a `while true do () done`, it would run forever on a worker, potentially preventing the execution of some other jobs that might terminate faster (or at all).

ElectreAAS:
Resolves #392. The duration of the timeout is for now 2 hours long but that's up to discussion of course.


punchagan (approved):
Thanks for identifying that we need a timeout, and adding it!

PR #398, opened 2022-11-21
Upper bound on dune < 3.6 because it breaks obuilder

ElectreAAS:
As per [the release notes](https://github.com/ocaml/dune/releases/tag/3.6.0), dune doesn't allow some private library shenanigans that happen in obuilder. This PR fixes that and allow current-bench to actually build.


art-w:
Merci !

PR #402, opened 2022-11-29
Fix broken GitHub link

### LINKED ISSUE #382: Github link button doesn't work
When clicking the Github button here
image
The link isn't followed and nothing happens. Console says "Uncaught DOMException: The operation is insecure."

punchagan:
Closes #382

### DIFF
--- a/frontend/src/App.res
+++ b/frontend/src/App.res
@@ -386,10 +386,12 @@ module RepoView = {
               })}
             </Row>
           let githubLink =
-            <Link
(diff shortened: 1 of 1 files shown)


ElectreAAS:
Looks almost good to me :)
My only qualm would be that the icon looks a little greyed out, is that intentional?
image

punchagan:
> My only qualm would be that the icon looks a little greyed out, is that intentional?

I made it look similar to the greyed out GitHub icons in the sidebar. But, I guess it makes sense to change that.

ElectreAAS (approved):
LGTM :)

PR #405, opened 2023-01-05
Replace deprecated Mtime.Span.to_ms with to_float_ns /. 1e6

ElectreAAS:
We pull mtime version 2.0.0 with some of our dependencies.
`Mtime.Span.to_ms` has been removed in version 2.0, and using `to_float_ns` seems to be the replacement. This should fix the CI


ElectreAAS:
Other PRs should rebase onto this one when it's merged

punchagan (approved):
Thanks for the quick fix!

PR #407, opened 2023-01-23
Fix url for http endpoint in README

ElectreAAS:


README.md
```diff
@@ -69,7 +69,7 @@ Benchmarks data could also be added directly to the DB without having
 `current-bench` running the benchmarks using a HTTP end-point.
```
art-w, line 69: Perhaps add a hint that one should request a `token` password for their repo?

art-w (approved):
Thanks!

README.md
ElectreAAS: Yes that would make sense, but I don't know who would be the point of contact for that. And I'm afraid it would be me, and I don't know how to give someone a `token`.

art-w: Yes that could likely be you! But I'm very confident that you would be able to find out how to give them a new token after a quick grep :P (I'm half kidding by not giving you the answer straight away, but we can look at it together after you give it a try :) )

PR #410, opened 2023-02-02
Use literal value in json format example in readme

gridbugs:
I imagine there is a tendency to copy this example when making one's first benchmark, and removing the obvious placeholders (the ellipses). The problem with including `"value": "benchmark-value"` in the format example is that it is valid json but isn't accepted by current-bench, so it may take a while for people to realize that this is the reason their benchmarks aren't producing graphs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ To be able to draw graphs from your results, they need to follow this format:
     {
       "name": "name-of-the-test",
       "metrics": [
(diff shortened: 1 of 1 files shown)


ElectreAAS:
I didn't even know this could cause a problem. Nice catch and thanks for the PR!

ElectreAAS (approved):

PR #412, opened 2023-02-09
Fix validation for OCAML_BENCH_DOCKER_CPU env variable

punchagan:
This commit the OCAML_BENCH_DOCKER_CPU validation to allow both comma-separated
CPU numbers and comma-separated CPU ranges. We also ensure that the number of
CPUs in each of the ranges is the same.


art-w:
Thanks!

PR #413, opened 2023-02-13
Checking docker runs before trying to kill it

ElectreAAS:
This removes a slightly annoying warning:
 `worker_1 | Error response from daemon: Cannot kill container: cb_worker_runner_0: No such container: cb_worker_runner_0`
We first check if `docker ps` finds that container before `kill`ing it.

### DIFF
--- a/worker/cb_worker.ml
+++ b/worker/cb_worker.ml
@@ -129,10 +129,12 @@ let docker_run ~switch ~log ~docker_config ~cpu img_hash =
     @ run_command ~arch:docker_config.arch
   in
   Lwt_switch.add_hook_or_exec (Some switch) (fun () ->
-      Lwt_unix.system ("docker ps -q | grep -q " ^ name) >>= function
-      | Unix.WEXITED 0 ->
-          Lwt_unix.system ("docker kill " ^ name) >>= fun _ -> Lwt.return_unit
-      | _ -> Lwt.return_unit)
+      Lwt_process.pread ("docker", [| "docker"; "ps"; "-qf"; "name=" ^ name |])
+      >>= function
+      | "" -> Lwt.return_unit
+      | _ ->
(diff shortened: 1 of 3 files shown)


worker/cb_worker.ml
```diff
@@ -129,7 +129,10 @@ let docker_run ~switch ~log ~docker_config ~cpu img_hash =
     @ run_command ~arch:docker_config.arch
   in
   Lwt_switch.add_hook_or_exec (Some switch) (fun () ->
-      Lwt_unix.system ("docker kill " ^ name) >>= fun _ -> Lwt.return_unit)
+      Lwt_unix.system ("docker ps -q | grep -q " ^ name) >>= function
```
art-w, line 132: Perhaps [`docker ps --filter ...`](https://docs.docker.com/engine/reference/commandline/ps/#filter) can be used instead of grep? :) (and then perhaps with `Lwt_process` to avoid any quoting issues related to executing the command in a shell?)

ElectreAAS: Thanks for the review. I wanted to have a dumb but "visually correct" solution out before I forgot about it. I'll look into doing it correctly later today

ElectreAAS: I am pretty sure `Lwt_unix.system "foo bar"` is equivalent to `Lwt_process.exec (Lwt_process.shell "foo bar")`.
If we don't use `Lwt_process.shell` (to avoid the shell itself), we need to construct the value manually: `Lwt_process.exec ("foo", [|"foo"; "bar"|])` which is quite unwieldy, but maybe it is more correct?

ElectreAAS: @art-w The newest version should be cleaner

art-w: Thanks!

Issue #395, opened 2022-11-10
Allow bench.Dockerfile to exist outside the root


ElectreAAS:
Having a `bench.Dockerfile` at the root of the project isn't very clean, especially for big projects. It would be nice to unclutter this by allowing the file to be anywhere, or a specific other place..

art-w:
A simple solution would be to list the custom dockerfile path/filename in the repo configuration(s), as we already do for the docker images, which would then also enable having multiple dockerfiles in different environments (as requested by @gridbugs )

PR #414, opened 2023-02-15
Allow dockerfiles to exist anywhere, and there can be multiple ones

ElectreAAS:
This PR solves #395.
Users can tell us where their dockerfile is, and we'll setup a `"dockerfile": "path/to/dockerfile"` in `production.conf` and it should be good to go.

If they want to have multiple dockerfiles, it's now possible:
```json
{
  "name": "org/repo",
  "dockerfile": "path/to/normal/dockerfile"
},
{
  "name": "org/repo",
  "dockerfile": "path/to/other/dockerfile"
}
```
Tested locally and everything seems to be in order.

### DIFF
--- a/pipeline/lib/config.ml
+++ b/pipeline/lib/config.ml
@@ -9,6 +9,7 @@ type repo = {
   name : string;
   worker : string; [@default default_worker]
   image : string; [@default default_docker]
+  dockerfile : string option; [@default None]
   schedule : string option; [@default None]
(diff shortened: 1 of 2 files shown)


art-w:
Very nice, thank you! We should update the readme with the list of extraordinary settings that one can ask for when integrating with current-bench :) (custom dockerfiles, ocaml versions, single vs multicore, github labels, ...)

PR #416, opened 2023-02-27
Proof of concept of CB-checker

ElectreAAS:
This PR creates two new packages: `cb-schema` which is used internally as the one-stop API, and `cb-check` which is an executable taking in a file or stdin and checking that the produced json is of the correct format.

This would probably solve #385.

PS: I unified the various `.ocamlformat` files that were lying around for some reason, sorry for the noise


ElectreAAS:
This is in draft format for the moment, because I can't get the packaging right, but the main code is ready for review @art-w @punchagan !

cb-check/cb_check.ml
```diff
@@ -0,0 +1,52 @@
...
+                  s
+                  (Yojson.Safe.pretty_print ~std:false)
+                  j;
+                None)
+        | exception Yojson.Json_error s ->
+            Format.eprintf
+              "\x1b[91mJson parsing failure, please report: \x1b[0m%s" s;
+            None)
+  in
+  match loop l [] with
```
art-w, line 35: Perhaps the `loop` can be replaced by a `List.filter_map`? (It's not technically a blocking error if one of the detected json fails, but it is nice to report it as a warning! The real error is if no schema-valid jsons are detected at all by the end)

cb-check/cb_check.ml
```diff
@@ -0,0 +1,52 @@
...
+  | None -> ()
+  | Some l ->
+      let merged = Cb_schema.S.merge [] l |> List.map Cb_schema.S.to_json in
+      Format.printf "Correctly parsed %d benchmark(s):\n" (List.length merged);
```
art-w, line 39: Nitpick: we may have parsed more valid jsons than what remains after the merge (which could collapse everything down to one result :P )

cb-check/cb_check.ml
```diff
@@ -0,0 +1,52 @@
...
+      List.iter
+        (fun j -> Format.printf "%s@;" (Cb_schema.Json.pp_to_string () j))
+        merged
+
+let () =
+  let ic = if Array.length Sys.argv >= 2 then open_in Sys.argv.(1) else stdin in
```
art-w, line 45: As a later improvement, it might be worth looking into `Cmdliner` (only for its ability to generate a user friendly `--help` man page, we don't need crazy command line arguments... for now :D)

ElectreAAS:
The new commit history should be much clearer, and allow for more precise reviewing.
In the current state, the project doesn't build due to, I assume, linking problems at runtime (cb-schema not being available in dockers probably)

cb-check/cb_check.ml
ElectreAAS: If I understood correctly, this should be fixed now

cb-check/cb_check.ml
ElectreAAS: Do you mean that in some cases, even though a merge would succeed, it would not be desirable? Or that we need to treat the result of the merge as a list?

cb-check/cb_check.ml
art-w: Yes it looks good! Thanks :)

ElectreAAS:
As #422 is now merged, the content is upstreamed, and with #424 the noise of the packaging has been resolved. I'll close this PR and associated issues as completed.

PR #417, opened 2023-02-28
Dependency fixes

ElectreAAS:


ElectreAAS:
The pin changes and mirage crypto stuff is the same as in this commit of ocurrent: https://github.com/ocurrent/ocurrent/commit/c611b108dbb9af2611a34dc57e13eb2ce5af886d

PR #422, opened 2023-03-09
Cb check standalone

ElectreAAS:
This PR takes on the important parts of #416, but without the hassle of exposing `cb-schema` as a public toplevel library, which it turns out completely messes up our docker setup.
Instead, it is now a library inside the pipeline, which is properly linked to and accessed when needed.

I also incorporated this comment: https://github.com/ocurrent/current-bench/pull/416#discussion_r1121924894


dune
```diff
@@ -0,0 +1 @@
+(dirs :standard \ frontend)
```
ElectreAAS, line 1: This allows running a simple `dune build` from the root of the project, without having errors related to something in `frontend/node_modules/foo/bar/deep/deeper/something.ml`

pipeline/Dockerfile.dev
```diff
@@ -20,7 +20,7 @@ WORKDIR /mnt/project
 
 # Build dependencies.
 COPY --chown=opam:opam pipeline/pipeline.opam pipeline.opam
-RUN opam install -y --deps-only -t .
+RUN opam install -y --deps-only .
```
ElectreAAS, line 23: installing with the `-t` flag makes it longer without a good motive. In the prod Dockerfile we already don't use it

pipeline/dune
```diff
@@ -0,0 +1 @@
+(dirs :standard \ var)
```
ElectreAAS, line 1: Similar to the toplevel dune file, this is necessary as with multiple packages defined in `pipeline/dune-project`, dune checks all directories, including `pipeline/var` which is the ocurrent working directory.

cb-check/main.ml
```diff
@@ -0,0 +1,52 @@
...
+  match List.filter_map aux l with
+  | [] ->
+      Format.printf "No schema-valid results were parsed.\n";
+      exit 1
+  | validated ->
+      print_endline "Correctly parsed some benchmarks:";
+      Cb_schema.S.merge [] validated
+      |> List.map Cb_schema.S.to_json
+      |> List.iter (fun j ->
+             Format.printf "%s@;" (Cb_schema.Json.pp_to_string () j))
```
art-w, line 41: Nitpick: using `@.` for line breaks might work better with `Format` since it also forces a flush (while `@;` may choose not to break but [output spaces](https://v2.ocaml.org/api/Format.html#VALprint_break), and in above lines you use `\n` which doesn't cause Format to flush its internal buffer to sdout). It would also be nice to only use Format rather than mix it with `print_endline` (again, for buffering reasons)

pipeline/cb-schema/json.ml
```diff
@@ -0,0 +1,122 @@
...
+       foo: [
+         bar
+       ]
+     rather than something else like
+       foo:
+         [
+           bar
+         ]
+  *)
+let rec format ~inside_box out (x : t) =
```
art-w, line 44: Wasn't it possible to use the yojson's formatters? (I understand that the frontend can't use yojson, but then it also doesn't need to pretty-print the jsons anyway?)

cb-check/main.ml
ElectreAAS: According to https://v2.ocaml.org/releases/5.0/api/Format.html#VALprint_newline, `@.` is equivalent to `print_newline`, which isn't recommended. Do you think we should break from their recommendations?
I agree that the `print_endline` is misplaced though

pipeline/cb-schema/json.ml
ElectreAAS: The frontend only directly uses `json.ml` to refer to the type `t`, but it also uses some functions in `schema.ml` which in turn call functions in `json.ml`, including some that can error out, with the `error` function (which uses the pretty-printers)

I don't know how to solve this cleanly, though I agree that copy-pasting Yojson code isn't the best

ElectreAAS: One possible option would be to ugly_print in cb-schema, and to pretty_print in cb-check (or even use `jq` as it is even prettier)

ElectreAAS: > One possible option would be to ugly_print in cb-schema, and to pretty_print in cb-check

This is what I ended up doing, and it is much cleaner indeed

cb-check/main.ml
art-w: Yes I think so! The Format documentation was written with the intent that one would use Format to pretty-print large ASTs, and in that context, using `@.` is wrong since it signals the end of the document (to be flushed on stdout) rather than a recommended line-break in the syntax of the AST. But when using [Format as a better Printf](https://sherlocode.com/?q=%40.%22) it makes sense to output small lines where `@.` is perfectly appropriate to end the print :) (we want Format to output immediately, and the following lines should not be influenced by an earlier formatting bug, for example a non-closed `@[` leading to absurd indentation)

pipeline/cb-schema/json.ml
art-w: Ha ok, I now see that extracting informations from a json uses this formatting to provide debugging information on a type error. (My original suggestion was to remove all json pretty-printing code and use `Yojson.Safe.pp..` in place of `Cb_schema.Json.pp` where needed in `cb-check` and the `pipeline`... This could still be done by having the `error` function raise a specific exception catchable by whomever is interested in pretty-printing json, but it's not that important)

ElectreAAS:
As the CI passes and the error reporting has been improved, I now declare `cb-check` to be in beta!
I'll incorporate user-suggested changes and merge when satisfied

ElectreAAS:
The force-pushes were rebases and a clean-up of the commit history, without changes. As everyone seems happy, I'll merge as-is.

PR #424, opened 2023-03-16
Nitpicking & Cleaning

ElectreAAS:
This PR adds, removes, or changes no functionality, it just aims at making the development process easier.
A new dev can just `opam install . --deps-only --with-test` and `dune build` will just work! Without warnings!

Rationale for things I *didn't do*:
- Completely unify `dune-project` files. This can't be done because the pipeline and the worker boot their docker environments from their respective directories, and try to build in them. This requires a `dune-project` for both.
- Completely unify `.ocamlformat` files. This follows from the previous constraint. OCamlformat detects projects with the `dune-project` files, hence we need a new one for the worker and the pipeline.
- Unify `dune` (with `dirs :standard...`). This can't be done in a single `dune` file at the root as of dune 3.7.0, we need one at each level.
- Add lower bounds to all dependencies. This would require a lot of effort to not break dependency chains, and the CI is happy as is.


dune
```diff
@@ -0,0 +1 @@
+(dirs :standard \ frontend)
```
ElectreAAS, line 1: This allows running a simple `dune build` from the root of the project, without having errors related to something in `frontend/node_modules/foo/bar/deep/deeper/something.ml`

pipeline/Dockerfile.dev
```diff
@@ -20,7 +20,7 @@ WORKDIR /mnt/project
 
 # Build dependencies.
 COPY --chown=opam:opam pipeline/pipeline.opam pipeline.opam
-RUN opam install -y --deps-only -t .
+RUN opam install -y --deps-only .
```
ElectreAAS, line 23: Installing with the `-t` flag makes it longer without a good motive. In the prod Dockerfile we already don't use it.

punchagan (approved):
Thanks for taking the time to do this! :heart:

PR #425, opened 2023-03-21
Moving to latest dune

ElectreAAS:
Since #420, if building using the latest versions of dune (above `3.5.0`), we had an error in the pipeline:
```
Error: A running dune (pid: 861738) instance has locked the build directory.
If this is not the case, please delete _build/.lock
```
This is because in `pipeline/reload.sh` we use `dune build --watch` **and then** call `dev/github-app.sh`, which calls `dune build`. This workflow has been made impossible in dune 3.6 and up with the use of lockfiles.

The simplest fix was to use `dune exec --watch`, introduced in the latest version of dune (`3.7.0`), and to call `dev/github-app.sh` **before** the call to watch.
Using the latest dune version introduced a bit of noise, sorry for that.

### DIFF
--- a/pipeline/reload.sh
+++ b/pipeline/reload.sh
@@ -1,33 +1,8 @@
 #!/bin/bash
+set -eu
 
-COMMAND="${*@Q}"
+cd /mnt/project
 
-DIR=_build/default/bin
-EXE="${DIR}/main.exe"
+./dev/github-app.sh
 
-function run {
-  cd /app
-  sh -c "$COMMAND" &
-  cd /mnt/project
-}
-
-run
-
-dune build --watch bin/main.exe &
-
-/mnt/project/dev/github-app.sh
-
-while :
-do
-  # NOTE:
-  # 1. inotifywait exits with a failure status code when main.exe gets deleted
-  #    before a rebuild. So, we watch the directory instead, and make sure the
-  #    file exists.
-  # 2. Also, we sleep for a short while, to ensure the process has been killed
-  #    before copying the new executable and trying to run it.
-  inotifywait -q -e CLOSE_WRITE "${DIR}" | grep -q main.exe \
-  && (pkill -f '^/app/bin/current-bench-pipeline' || echo 'Not running?') \
-  && sleep 0.1 \
-  && cp "${EXE}" /app/bin/current-bench-pipeline \
-  && run
-done
+dune exec --watch bin/main.exe -- "$@"
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ frontend/node_modules
 current-bench-bechamel
 scripts
 capnp-secrets
+pipeline/_build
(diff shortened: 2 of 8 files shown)


pipeline/reload.sh
ElectreAAS, line 1: Special reviewing attention should be paid to these changes as shell quoting is notably difficult.
The arguments to this script are located in `development.docker-compose.yaml`, and look like `"--foo=bar" "--baz" "--spam=eggs break=fast"`. This list was prepended with the executable produced by dune to call it directly.
We used to do `COMMAND="${*@Q}"`, then `sh -c "$COMMAND"`.

What I now do is remove the direct call to the executable, in favor of a simpler `dune exec bin/main.exe -- "$@"`. This means the arguments now need not be super quoted because of the call to `sh`, but still need to be somewhat quoted so they don't break dune.

punchagan:
> This is because in `pipeline/reload.sh` we use `dune build --watch` **and then** call `dev/github-app.sh`, which calls `dune build`. This workflow has been made impossible in dune 3.6 and up with the use of lockfiles.

Thanks for working on fixing this! 

I also see an error that goes like this, when I stop the development docker container and restart it. 

```
error checking context: no permission to read from '/home/punchagan/code/segfault/current-bench/pipeline/_build/.lock'
```

Adding `pipeline/_build` to the `.dockerignore` file helps. Would that be useful to add?

punchagan:
> This is because in `pipeline/reload.sh` we use `dune build --watch` **and then** call `dev/github-app.sh`, which calls `dune build`. This workflow has been made impossible in dune 3.6 and up with the use of lockfiles.

Thanks for working on fixing this! 

I also see an error that goes like this, when I stop the development docker container and restart it. 

```
error checking context: no permission to read from '/home/punchagan/code/segfault/current-bench/pipeline/_build/.lock'
```

Adding `pipeline/_build` to the `.dockerignore` file helps. Would that be useful to add?

ElectreAAS:
> I also see an error that goes like this, when I stop the development docker container and restart it. `error checking context: no permission to read from '/home/punchagan/code/segfault/current-bench/pipeline/_build/.lock'`

I can't reproduce that... But I `rm -rf pipeline/_build` several times recently to clean up my directories, maybe that's the cause? If doing that doesn't solve the error, could you give me more details?

punchagan:
> I can't reproduce that... But I `rm -rf pipeline/_build` several times recently to clean up my directories, maybe that's the cause? If doing that doesn't solve the error, could you give me more details?

If I remove the files, the docker container starts up, but I run into the issue again if I shutdown the container and start it again. It may be a difference between running docker on Linux vs MacOS hosts, which is why you're probably not running into it.

ElectreAAS:
I added it to the ignore file. As I can't test it on my machine I'm going to trust you on this one :)

punchagan (approved):
Looks good to me! Thanks!

PR #427, opened 2023-03-30
Support using a separate benchmarks repo for benchmarks

punchagan:
For repositories like ocaml/ocaml, we would like the benchmarks to be maintained potentially by a bigger/different set of users, than the ocaml/ocaml maintainers. This commit adds support for an additional configuration variable that lets us clone an additional repository and run benchmarks from within it.

NOTE: No new benchmark runs would be triggered by any changes to the benchmarks repository, with this setup.

This is a simplistic PR that attempts to implement functionality similar to #370 without the UI hackery.

### DIFF
--- a/environments/development.conf
+++ b/environments/development.conf
@@ -3,7 +3,8 @@
     {
       "name": "local/test",
       "worker": "autumn",
-      "image": "ocaml/opam:debian-11-ocaml-4.14"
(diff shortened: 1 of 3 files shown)


ElectreAAS:
Wow indeed this is a very elegant solution! Bravo!
I'll test it myself, and if everything runs fine I'll merge :)

ElectreAAS (approved):
LGTM!

PR #428, opened 2023-04-05
Filter out branches for benchmarking before setting up metadata

punchagan:
Previously, we filtered out the branches which don't have the required label, before starting the benchmarking run. But, this meant that the metadata for the branch is already saved to the DB and displayed in the UI. This commit changes the filtering to be run before the metadata for the benchmark run is setup.


ElectreAAS:
Closing this as #430 incorporates the changes here

PR #429, opened 2023-04-11
Specifically pin submodules in ocluster

ElectreAAS:
According to @MisterDA, this is necessary to properly use the vendored submodules.
Without it, a build with the latest docker pulls fails because of an "undefined module ..."


art-w (approved):
Looks good!

PR #430, opened 2023-04-13
Support running benchmarks on multiple long-standing branches of a repository

### LINKED ISSUE #426: Support specifying additional branches to run benchmarks on
Currently, only the default_branch and PRs are considered when running the benchmarks. It would be useful to specify different tags/branches on which to run the benchmarks. It would be useful to run benchmarks against different versions of development like OCaml 5.0.x and 5.1.x, for instance.

punchagan:
Previously, we only ran the benchmarks on the default branch of a repository, apart from the pull requests. This PR adds support to run benchmarks on multiple branches of a repository, not just the default branch.

NOTE: This also includes the change in #428. That PR can either be merged before this one, or simply closed, in favor of this.

Closes #426

### DIFF
--- a/pipeline/db/migrations/20230406110951_add_pull_base_column.up.sql
+++ b/pipeline/db/migrations/20230406110951_add_pull_base_column.up.sql
@@ -1,0 +1,39 @@
+ALTER TABLE benchmark_metadata ADD pull_base varchar(100);
+ALTER TABLE benchmarks ADD pull_base varchar(100);
+
+WITH default_branches (repo_id, pull_base) AS (VALUES
+('odis-labs/streaming', 'master'),
+('art-w/ocaml-benching', 'master'),
+('gridbugs/dune-monorepo-bench', 'main'),
+('Zineb-Ada/merlin', 'master'),
+('ocaml-multicore/kcas', 'main'),
+('mirage/irmin', 'main'),
+('ocaml-community/yojson', 'master'),
+('art-w/cb-dev-test', 'master'),
+('punchagan/current-bench', 'main'),
+('mirage/repr', 'main'),
+('mirage/index', 'main'),
+('ocaml-ppx/ppxlib', 'main'),
+('Zineb-Ada/bechamel-fact', 'master'),
+('ocaml-bench/sandmark', 'main'),
+('ocaml/dune', 'main'),
+('ElectreAAS/sandmark', 'main'),
+('ocaml-multicore/lockfree', 'main'),
+('Lucccyo/cachecache', 'main'),
+('ocaml-ppx/ocamlformat', 'main'),
+('art-w/ocaml', 'cb-comanche'),
+('Zineb-Ada/eqaf', 'master'),
+('ocaml/merlin', 'master'),
+('ElectreAAS/Paradict', 'master')
+),
+update_metadata AS (
+  UPDATE benchmark_metadata
+  SET pull_base = default_branches.pull_base
+  FROM default_branches
+  WHERE benchmark_metadata.repo_id = default_branches.repo_id
+  RETURNING benchmark_metadata.repo_id
+)
+UPDATE benchmarks
+SET pull_base = default_branches.pull_base
+FROM default_branches
+WHERE benchmarks.repo_id = default_branches.repo_id;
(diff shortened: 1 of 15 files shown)


pipeline/lib/pipeline.ml
```diff
@@ -181,13 +181,29 @@ let github_repositories repo =
...
+          let branch = Util.get_branch_name branch in
+          let repository = repository ~branch ~labels:[] () in
+          match
+            List.filter
+              (Config.must_benchmark_branch ~default_branch:default_branch_name
+                 ~branch)
+              (Config.find config repository)
+          with
+          | [] -> lst
+          | _ -> repository :: lst)
```
ElectreAAS, line 194: Not quite sure about this bit. Why drop the result of the filter instead of adding it to `lst`?
Same question for the `PR` case down

ElectreAAS (commented):
Overall the code seems good (bar my minor comments), but I am not able to properly test that it works.
I tried adding
```json
{ "name": "ElectreAAS/merlin",
  "branches": [ "master", "mock-without-label" ]
}
```
to `environment/development.conf`, and adding commits to my mock repo, but the only thing that seems to run is the `master` branch.
What would be a better way to test it?

pipeline/db/migrations/20230406110951_add_pull_base_column.up.sql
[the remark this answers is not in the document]
punchagan: The query is only updating existing entries in the DB. The pipeline won't do anything for PRs already existing in the DB, but the UI would need to know what the base branches for these PRs are. For future PRs, this information would be automatically populated.

pipeline/lib/pipeline.ml
punchagan: The result of the filter is a list of config objects, where the current branch (or PR in the case below) returned by the API is configured to be benchmarked. In the case, where a whitelisted configuration exists for the current branch, we add the repository object to the list of repository objects for benchmarking.

punchagan:
> Overall the code seems good (bar my minor comments), but I am not able to properly test that it works. I tried adding
> 
> ```json
> { "name": "ElectreAAS/merlin",
>   "branches": [ "master", "mock-without-label" ]
> }
> ```
> 
> to `environment/development.conf`, and adding commits to my mock repo, but the only thing that seems to run is the `master` branch. What would be a better way to test it?

This is how I tested it, too. Did you restart the containers after changing the `development.conf` file? Also, we ignore commits which are older than 2 weeks old, when running benchmarks (See `Pipeline.filter_stale_repositories`).  I can't think of any other reason why you're unable to test it.

punchagan:
> Overall the code seems good (bar my minor comments), but I am not able to properly test that it works. I tried adding
> 
> ```json
> { "name": "ElectreAAS/merlin",
>   "branches": [ "master", "mock-without-label" ]
> }
> ```
> 
> to `environment/development.conf`, and adding commits to my mock repo, but the only thing that seems to run is the `master` branch. What would be a better way to test it?

This is how I tested it, too. Did you restart the containers after changing the `development.conf` file? Also, we ignore commits which are older than 2 weeks old, when running benchmarks (See `Pipeline.filter_stale_repositories`).  I can't think of any other reason why you're unable to test it.

pipeline/db/migrations/20230406110951_add_pull_base_column.up.sql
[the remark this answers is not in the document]
ElectreAAS: Oh ok cool

pipeline/lib/pipeline.ml
ElectreAAS: Ah I see. In that case I think I'd prefer something along the lines of
```ocaml
if List.exists
  (Config.must_benchmark_branch ~default_branch:default_branch_name
    ~branch)
  (Config.find config repository)
then repository :: lst
else lst
```
To make it clear that the only info we want is "is there at least one element in the list that matches the filter?"

ElectreAAS:
There was a CI failure that vanished when I rebuilt it, I have no clue what caused it.
I'll consider this good to go.

punchagan:
Thanks for the careful review and merge, @ElectreAAS !

PR #431, opened 2023-04-19
Couple of improvements to using a separate bench_repo

punchagan:
1.Skip installing opam dependencies of the repo when using separate `bench_repo`
2. Ensure cached `bench_repo` is used only when the `bench_repo` has no changes. We use the GitHub refs API to see if the repo has changed, to decide whether to use the cached bench_repo or make a new clone of it.  The trick here is from this GitHub [issue comment](https://github.com/moby/moby/issues/1996#issuecomment-1152463036).
3. Support specifying a specific branch of the `bench_repo` to use for running the benchmarks.  The branch can be specified using the `/tree/<branch-name>` suffix in the bench_repo URL.

### DIFF
--- a/pipeline/lib/custom_dockerfile.ml
+++ b/pipeline/lib/custom_dockerfile.ml
@@ -111,15 +111,11 @@ let dockerfile ~base ~files ~bench_repo =
   @@ run "opam update"
   @@ run "mkdir bench-dir && chown opam:opam bench-dir"
   @@ workdir "bench-dir"
-  @@ install_opam_dependencies ~files
   @@ copy ~chown:"opam:opam" ~src:[ "." ] ~dst:"." ()
-  @@ (match (bench_repo, bench_dir) with
-     | Some repo, Some dir -> run "git clone %s %s" repo dir
-     | _ -> comment "No bench repo to clone")
   @@
-  match bench_dir with
-  | Some dir -> workdir "%s" dir
-  | _ -> comment "No extra bench repo"
+  match (bench_repo, bench_dir) with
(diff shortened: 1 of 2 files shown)


pipeline/lib/custom_dockerfile.ml
```diff
@@ -95,12 +95,34 @@ let install_opam_dependencies ~files =
       @@ run "opam exec -- opam install -y --depext-only %s" targets
       @@ run "opam exec -- opam install -y --deps-only --with-test %s" targets
 
+let contains_at s1 s2 =
+  let re = Str.regexp_string s2 in
+  try
+    let idx = Str.search_forward re s1 0 in
+    idx
+  with Not_found -> -1
```
ElectreAAS, line 103: Nitpick: I'd rather use `int option` here than using the somewhat obscure (and not type-safe) `-1` to signal something went wrong.

ElectreAAS (approved):
Aside from my nitpicking, everything looks good to me.
I think we can ship with the FIXME regarding github assumption, and add an issue to `git clone` properly from other sources afterward
Thanks!

pipeline/lib/custom_dockerfile.ml
punchagan: Thanks for catching this. I've pushed a change to fix this.

punchagan:
> I think we can ship with the FIXME regarding github assumption for the time being, and add an issue to `git clone` properly from other sources afterward

Yes, I agree. That was my intention too, with adding the FIXME.

punchagan:
Thanks for the review and the merge, @ElectreAAS!

PR #437, opened 2023-05-02
Use latest ocurrent to fetch most recent PRs and branches

punchagan:


ElectreAAS:
Thanks for both this PR and the one in ocurrent. I'll put in some minor edits to clarify the pin, but it seems good to me

PR #438, opened 2023-05-02
Nitpicking: using infix operator combines nicely into options

ElectreAAS:
In #434 I mentioned that
> We could use options instead of exceptions or fancy `let*` syntax or such

Well here's an attempt to make it look nicer.
Feel free to make stylistic remarks


art-w (approved):
It's a bit gratuitous but why not :P

PR #444, opened 2023-05-16
Handle 0 values correctly when adjusting units for sizes

punchagan:


ElectreAAS (approved):
Ah, the nasty `log10(0) = -inf`... 
Nice catch, thanks for the PR!

PR #455, opened 2023-06-12
Ensure new builds are triggered on pushes to the `main` branch

punchagan:
With the current app settings, new builds are not triggered when a new commit is pushed to `main`. Pushes to PR branches do trigger a `POST` to the webhook, and `current-bench` ends up checking all the branches for updates, which ensures that the latest commit on `main` is benchmarked too. But, triggers to `main` don't directly trigger a new build.

Adding read permission on `contents` and enabling webhook triggers for the `push` event would fix this problem.


punchagan:
The permissions have been changed on the production `ocaml-benchmarks` application, which would trigger emails being sent to existing users of the app.

art-w (approved):
Very nice catch! This has been pending for a long time, I'm super happy you found the solution :tada:

PR #460, opened 2023-06-19
Pin docker images

### LINKED ISSUE #459: Pin the docker images to specific versions
We have a whole bunch of Docker images we depend on, and most of them are just "lts" or "latest". To avoid surprises, it would be nice to pin them to specific versions. A bonus would be to have a tool to upgrade them all, if we choose to.

punchagan:
Pin versions of the docker images being used

Closes #459


ElectreAAS:
I agree that we need to do something along those lines, but I'd wager we go a step further and do it the ocaml-ci way:
- Have a specific hash of a docker image so we're sure it's stable
- Have a specific hash of the opam repository, so we know what versions of ocaml packages we're running on
(example here https://github.com/ocurrent/ocaml-ci/blob/master/Dockerfile#L1-L3)

And I think it's important that we put a note somewhere (I don't know where though) explaining the whole thing: "This hash was the latest when we modified this file, and we tested it to make sure it runs fine. Update this in X months."
And then actually follow through with it

punchagan:
Thanks these suggestions are great. I'll work on updating the PR.

I think having a simple tool to change these hashes might be helpful in making sure that we actually do the updates, every so often. It might be a useful thing for other teams like the CI team too. Maybe there's already something like this out there, otherwise might be a good Hacking days project.

punchagan:
Self-merging since this has been manually tested by @shakthimaan a couple of times at least.

PR #456, opened 2023-06-12
Fix the github-app.sh script that allows locally testing with a GitHub app

punchagan:
This PR fixes a couple of things that were broken which broke the local dev workflow with a test GitHub app.  

1) ngrok added more domain names to their list of domains, but we looked only for ngrok.io
2) `dune exec` cannot be run in a project, after a `dune exec --watch` has already started.

The commit messages have more explanations on the bugs and the fixes.

PR #461, opened 2023-06-23
Run the github-app.sh only if OCAML_BENCH_GITHUB_* env vars are set

punchagan:
Running `dune exec dev/jwt.exe` crashes with a cryptic `Fatal error: exception Sys_error("Value too large for defined data type")` when development.env has empty values for the OCAML_BENCH_GITHUB_* environment variables. This commit ensures that we try to run this command and the `./dev/github-app.sh` script only when a couple of these variables are set.


punchagan:
Minor change that improves local developer setup, that was broken with #456 . Self-merging.

PR #463, opened 2023-06-28
Improve the label and link of the current-bench check in the list of checks for a PR

### LINKED ISSUE #453: Improve support for multiple configurations for a repository
Currently, we can add multiple configurations for a repository to run different benchmarks (or to use different workers) for the same repository.  For instance, the current configuration allows dune to run two different sets of benchmarks using different Dockerfiles using the following configuration: 

```json
    {
      "name": "ocaml/dune",
      "worker": "fermat",
      "dockerfile": "bench/monorepo/bench.Dockerfile",
      "notify_github": true
    },
    {
      "name": "ocaml/dune",
      "worker": "autumn"
    }
```

But, the status updates on GitHub are made to the same app with only the default URL. This makes the status of the benchmark runs unclear when one of the benchmarks is still running, while the other has finished (in this case, the monorepo benchmark, which takes longer). 

GitHub Actions, for instance, add multiple status URLs for each different kind of CI run (for example, see this dune [PR](https://github.com/ocaml/dune/pull/7904)). We should do something similar where each configuration posts a different update to GitHub.

punchagan:
When a repo has multiple configurations for different benchmark runs, the checks list previously displayed only entry which could be confusing for the users especially when one of the runs was much longer than the other. This PR fixes the number of entries to match with the number of configurations for a repo, and fixes the links to go to the correct page.

Closes #453

NOTE: The PR has an unrelated commit that improves #461 to print a log message when the local GitHub app isn't running.

### DIFF
--- a/pipeline/lib/pipeline.ml
+++ b/pipeline/lib/pipeline.ml
@@ -33,16 +33,17 @@ let github_status_of_state url = function
   | Error (`Active _) -> Github.Api.Status.v ~url `Pending
   | Error (`Msg m) -> Github.Api.Status.v ~url `Failure ~description:m
 
-let github_set_status ~repository ~worker ~docker_image result =
+let github_set_status ~repository ~config ~worker ~docker_image result =
   match Repository.github_head repository with
   | None -> Current.ignore_value result
   | Some head ->
-      let status_url = Repository.commit_status_url repository in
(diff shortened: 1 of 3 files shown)


art-w (approved):
Very good improvement! The multiple environments were always confusing so it's much nicer to have multiple entries that each points to the right url :)

PR #464, opened 2023-06-28
Update cb-check URL in README

Julow:
Github doesn't accept unauthenticated ssh connections. Use the https url for cb-check.


punchagan:
Thanks @Julow !

PR #470, opened 2023-07-05
Fix broken GitHub notifications for changed metrics

punchagan:
The code to figure out whether or not we should send a notification was unnecessarily complex and buggy.  This commit simplifies the code, and fixes the bug which caused notifications to be not sent for repositories which didn't have a dockerfile configured in the `.conf` file, but had the default dockerfile (`bench.Dockerfile`) present in the repository.


art-w (approved):
Ha, much simpler indeed!

PR #474, opened 2023-07-17
Fix metrics url to point to the PR page, instead of the benchmarks main branch page

punchagan:


punchagan:
Self merging based on tests on a test repository.

PR #482, opened 2023-11-13
Bench old commits

### LINKED ISSUE #379: Missing benchmarks when pushing multiple commits at once
We only benchmark the latest known commit when a git repository is updated... but the git push could have added more than one commit at the same time.  If we don't benchmark the intermediate commits, then the graph timeline is harder to read as it doesn't reflect the true git history. 

As a bonus, this might help with retro-benchmarking: We currently have to wait for each benchmark to complete, rather than submitting multiple commits corresponding to the different versions in one go!

(related to the removal of old commits https://github.com/ocurrent/current-bench/issues/378 which looks a bit easier to handle first)

ElectreAAS:
Fixes #379.
- `pipeline/lib/commits.ml` contains the important part: the actual git command to run to traverse the commit history.
- `pipeline/libs/models.ml[i]` contains the necessary two new functions that query the database, to make sure we don't benchmark twice the same commit.

~~Note: This PR requires some changes to the API of ocurrent (namely this PR: https://github.com/ocurrent/ocurrent/pull/442), and hence is in draft mode to make sure it doesn't get merged before the PR on ocurrent is properly reviewed.
It is, however, ready to be reviewed in the current-bench context. The API changes in ocurrent are trivials from our point of view.~~

Edit: ocurrent/ocurrent#442 has been merged, this is ready for review

### DIFF
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -70,7 +70,7 @@ pin-depends: [
   # Using the changes in ocurrent/ocurrent/pull/421, remove the pin on next release (> 0.6.4)
   ["current_github.0.6.4" "git+https://github.com/ocurrent/ocurrent.git#32561e1f2b9cc387d098e0ee918cd6bd2ccd1d3b"]
   # Using the changes in ocurrent/ocurrent/pull/442, remove the pin on next release (> 0.6.4)
-  ["current_git.0.6.4"    "git+https://github.com/ElectreAAS/ocurrent.git#21e7007e2897d44535e2729b42c80dc1280c6fb1"]
+  ["current_git.0.6.4"    "git+https://github.com/ocurrent/ocurrent.git#8e0b9d4bb348b13df8696fe63feba303b9a476fd"]
 ]
 
 # System dependencies fail to install on these platforms
--- a/current-bench.opam
+++ b/current-bench.opam
@@ -66,9 +66,11 @@ build: [
 ]
 dev-repo: "git+https://github.com/ocurrent/current-bench.git"
 pin-depends: [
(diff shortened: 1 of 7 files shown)


current-bench.opam
```diff
@@ -66,9 +66,11 @@ build: [
 ]
 dev-repo: "git+https://github.com/ocurrent/current-bench.git"
 pin-depends: [
+  # When modifying the pins in this file, don't forget to also modify pipeline/pipeline.opam
   # Using the changes in ocurrent/ocurrent/pull/421, remove the pin on next release (> 0.6.4)
-  # When modifying this pin, don't forget to also modify pipeline/pipeline.opam
   ["current_github.0.6.4" "git+https://github.com/ocurrent/ocurrent.git#32561e1f2b9cc387d098e0ee918cd6bd2ccd1d3b"]
+  # Using the changes in ocurrent/ocurrent/pull/442, remove the pin on next release (> 0.6.4)
+  ["current_git.0.6.4"    "git+https://github.com/ElectreAAS/ocurrent.git#21e7007e2897d44535e2729b42c80dc1280c6fb1"]
```
art-w, line 73: Should the pins point to ocurrent rather than your repo? (if the changes have been merged)

art-w:
Thanks! This was a much awaited feature!

A couple of thoughts which prevents this from going live:
- Can we disable it for repos that use a cron schedule? (`ocaml/ocaml` runs its benchmarks every week, precisely to avoid benchmarking too many commits)
- Is the commit order preserved when plotted on a graph?
- How do we get rid of old commits that were removed? (e.g. following a PR rebase)

current-bench.opam
ElectreAAS: Indeed, good catch

PR #484, opened 2023-11-16
Forbid bugged version 2.1.1 of yojson to fix tests

ElectreAAS:
The latest version of Yojson is bugged. See https://github.com/ocaml-community/yojson/issues/171 for details.
I added a version constraint to the main opam file so that the CI will pass, and there shouldn't be problems when `opam install . --deps-only` on a fresh clone.


punchagan:
Merged in #486

PR #485, opened 2023-11-16
[API tests] Properly use alcotest API instead of asserts

ElectreAAS:
Using `Alcotest.check_raises` instead of doing the `try / with` ourselves is clearer in the code and in the errors we get when tests fail.
Using `Alcotest.testable` values instead of cramming everything in asserts also yields better error messages on test failure.


art-w (approved):
Thanks to you both!

